PR #1, opened 2021-04-15
Merge in https://github.com/aryan26roy/fastjet

jpivarski:
The first few months of this project were developed in @aryan26roy's personal repo, but now that the build procedure works, I'm moving everything here without history. (Some mistakes were made in the early history; it's best to start with a clean git.)

At the end of this PR, it should be ready to use for all further development.


docs/conf.py
```diff
@@ -0,0 +1,49 @@
...
+
+# Warning: do not change the path here. To use autodoc, you need to install the
+# package first.
+
+from typing import List
+
+# -- Project information -----------------------------------------------------
+
+project = "fastjet"
+copyright = "2021, Scikit-HEP"
```
eduardo-rodrigues, line 15:
```suggestion
copyright = "2021, The Scikit-HEP admins"
```

jpivarski: https://github.com/scikit-hep/cookie is generating this by default; if you prefer the shorter version, also update it there.

eduardo-rodrigues:
Hi @jpivarski, @aryan26roy, very nice to see this coming in 👍 .

docs/conf.py
eduardo-rodrigues: Thanks, have just made the suggestion to Henry.

henryiii:
Check-manifest requires a recursive checkout.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    - master
+    - main
+    - develop
+
+jobs:
+  pre-commit:
+    name: Format
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
henryiii, line 17:
```suggestion
    - uses: actions/checkout@v2
       with:
         submodules: recursive
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    - uses: actions/setup-python@v2
+    - uses: pre-commit/action@v2.0.0
```
henryiii, line 19:
```suggestion
    - uses: pre-commit/action@v2.0.2
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+  hooks:
+  - id: flake8
+    exclude: docs/conf.py
+    additional_dependencies: [flake8-bugbear, flake8-print]
+
+# - repo: https://github.com/mgedmin/check-manifest
+#   rev: "0.45"
+#   hooks:
+#   - id: check-manifest
+#     stages: [manual]
```
henryiii, line 48:
```suggestion
- repo: https://github.com/mgedmin/check-manifest
  rev: "0.45"
  hooks:
  - id: check-manifest
    stages: [manual]
```

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      # Offical actions have moving tags like v1
```
henryiii, line 9:
```suggestion
      # Official actions have moving tags like v1
```

Pretty sure this is my typo :)

jpivarski: These things are coming directly from scikit-hep/cookie, which is why I had overlooked recursive checkouts.

I'll apply all of these as a single commit on my side. Thanks!

henryiii:
You need submodules v2 to use the `submodules: recursive` syntax. `v1` called it something else. For `v2` you need `fetch-depth: 0` if you use setuptools_scm.

.github/workflows/ci.yml
jpivarski: Actually, Fastjet has a submodule within it, for SISCone, so I don't want to limit the recursive depth.

jpivarski:
> You need submodules v2 to use the `submodules: recursive` syntax. `v1` called it something else. For `v2` you need `fetch-depth: 0` if you use setuptools_scm.

If `setuptools_scm` is what sets version numbers, I've disabled it for this repo because we want the version numbers to correlate with FastJet version numbers. For instance, the first version of this package will be 3.3.4.0rc1, where the fourth digit is the Python interface version.

henryiii:
Yes, then you don't need `fetch-depth: 0` (it'll be slightly faster without it).

.github/workflows/ci.yml
henryiii: There are only two options. `submodules: true` checkout out one level, and `submodules: recursive` checks out all levels. `fetch-depth` refers to the git history, not submodules.

henryiii:
> Rename fastjet -> fastjet-core to prevent unintended import in Python.

One of the reasons I always use an `/extern` directory. :)

jpivarski:
I don't know what's happening with

```
2021-04-16T00:51:31.4192598Z   checking CGAL in /tmp/pip-req-build-5mdakdlc/CGAL-5.2.1... lib
2021-04-16T00:51:31.6587274Z   checking CGAL/Exact_predicates_inexact_constructions_kernel.h presence... no
```

when the same task has

```
2021-04-16T00:51:22.5850739Z   -rw-r--r-- 1 runner docker 1586 Apr 16 00:51 /tmp/pip-req-build-5mdakdlc/CGAL-5.2.1/include/CGAL/Exact_predicates_inexact_constructions_kernel.h
```

and exactly the same thing always succeeds ("`presence... yes`") on my computer. I'm going to merge this just in case it's something to do with the submodule version on main vs the submodule version on this branch (i.e. if it's getting the wrong fastjet-core).

jpivarski:
Oh, I couldn't merge because it was a draft. Well, wait for success anyway.

PR #22, opened 2021-05-13
changes

aryan26roy:


jpivarski:
It's usually a challenge to make PRs small, but you've been having no problem making tiny PRs. I think that might be because you're just making each PR one commit, and that's not the idea either. A PR should take a repo from one working state to another, so that the main branch is always usable by users, so it should encompass a full feature, which might take multiple commits (looking at the tests) to get right.

Some of my recent PRs have been one commit (or two, counting the pre-commit bot's clean-ups), but that's because they were small bug-fixes. In development, that wouldn't often be the case.

At the very least, the PR name shouldn't be "changes." The reason for this PR workflow is so that each merged commit in the main branch's history (using "squash and merge") is one well-defined feature, so that we can go back in history to a working state right before one feature or another working state right after that feature was added. Development steps aren't perfect, and that's why commits are grouped into PRs, so that you get a chance to iterate toward a new working state. That's also what this message board is for, so that you can ask for help or otherwise communicate with your collaborators on the best way to get to that new working state.

aryan26roy:
Hey, I'll to make my PRs bigger now as you have pointed out. As for the the name "changes" that was supposed to be a test for the new Ubuntu installation to test out the ssh key(hence the deleted branch). However, I should've done that test on some private repo. Will keep that in mind from now on.

jpivarski:
I missed the fact that this is "closed" and not "merged." Actually, I see no problem doing such tests on a public repo if they're temporary (i.e. a PR that doesn't get merged like this).

Although it's hard to change the message of a commit, it's easy to change the title of a PR. If this was a temporary test, perhaps the title should reflect that. As it is, it looks like it's generic, unspecified changes to the codebase. (That's what I thought, which is why I commented.)

I'm not a very strong adherent of "keeping git (and GitHub) history clean." Having a well-documented history is helpful when you're trying to go back and understand what went wrong (or right) at some point in the past—archaeological debugging—but that's not something that happens all the time. Therefore, I don't think that keeping the history well organized should take a large amount of your time, but it should be organized enough that on those rare occasions when you do need to find something, it's not impossible.

So I'll change the title of this PR from "changes" to "Ubuntu version test." If you can be more specific, go ahead and change the title again.

PR #23, opened 2021-05-14
added custom class as output

aryan26roy:


aryan26roy:
@jpivarski , the CI doesn't seem to be working, I have manually requested runs as well, doesn't seem to work.

jpivarski:
When did that start?

Oh, GitHub is having issues. You'll just have to wait.

https://www.githubstatus.com/

PR #44, opened 2021-06-29
removed c++ printouts

aryan26roy:


jpivarski:
Are you able to press "Enable auto-merge (squash)"? If so, that will save you some time in not having to wait for tests to finish.

aryan26roy:
> Are you able to press "Enable auto-merge (squash)"? If so, that will save you some time in not having to wait for tests to finish.

I can't find the option anywhere.

PR #45, opened 2021-06-29
Preparing deployment YAML for first release.

aryan26roy:


jpivarski:
You're an administrator now (should have been all along; my mistake). You should see the "enable auto-merge" button now, but more important is manually triggering the wheels.yml action to test that.

aryan26roy:
Yes the option is available now. What I did in this latest commit was to comment out the dependency installation that was happening in ci.yaml and add vector to the setup.cfg. However the wheels.yml still did not work.

jpivarski:
Check the whole sequence of steps in ci.yml and wheels.yml. Somehow, ci.yml can compile, so the compilation parts of wheels.yml should be made the same. The YAML list items are just doing Linux commands, so it should be readable even if you don't know the GitHub workflow definition language (I don't, either). You can remove the MacOS case from wheels.yml because you don't have a working MacOS example in ci.yml.

aryan26roy:
@jpivarski , adding the extra installation step to wheels.yaml did not work.

aryan26roy:
@jpivarski , I seem to have made the whole build process similar to the CI but it does not work. Is there something I missed?

jpivarski:
It's not finding CGAL. Do you see that in the log message? How did we make CGAL visible in ci.yml? Do you remember? I don't remember exactly without looking into it.

aryan26roy:
As far as I remember, it was supposed to be the "Install extra deps on Linux" step. That is what worked for me when I switched distros.

.github/workflows/wheels.yml
```diff
@@ -26,12 +26,16 @@ jobs:
...
 
     steps:
     - uses: actions/checkout@v1
       with:
-        submodules: true
+        submodules: recursive
+
+    - name: Install extra deps on Linux
+      if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get install -y libmpfr-dev libboost-dev
```
henryiii, line 38: No, the wheels are not built here. They are built in a docker container. You need to set:

```
CIBW_BEFORE_ALL_LINUX: "yum update && yum install -y libmpfr-devel libboost-devel"
```

(or whatever these are called in CentOS 5 or 6, depending on if you are using manylinux1 or 2010). For macOS, you'll need to use brew instead.

henryiii: You might like the new pyproject.toml configuration in cibuildwheel 2.0.0.a4 instead, the final version will likely be out in a few weeks. That would be:

```toml
[tool.cibuildwheel.linux]
before-all = [
    "yum update",
    "yum install -y libmpfr-devel libboost-devel",
]
```

jpivarski: Thanks for the help! We've turned off MacOS in the wheels.yml because it's not being tested in ci.yml. We intend to turn that back on someday, but the current effort is for getting a usable first version as a release candidate on PyPI.

aryan26roy: Got it.  Is the command in the right place now? @henryiii

henryiii: Yes. :)

aryan26roy: I have corrected the names for CentOS, and the dependencies were installed as well (as per the log). What if the Fortran compiler is also an issue here? Because last time adding these dependencies worked for CGAL in Ubutnu. @henryiii

henryiii: I'm tempted to move this to cibuildwheel 2.0.0a4 and use pyproject.toml - then we can very simply add a test job building a single wheel, like I did for awkward (though we can do it already if we don't mind duplicating the setup variables - but not ideal). To test this currently you'll need to manually trigger it in the GitHub UI.

As long as you don't link to anything you can't bundle that's not in the manylinux list, you should be fine. The old fortran compiler should be just fine, I'd guess.

aryan26roy: Ah my bad, the error specifically mentions that it's a CGAL error, so it can't be a gfortran issue. But that also means that I have tried everything and it has not worked.

jpivarski: This is a modern project with a minimum Python of 3.6 (no 2.7). I doubt the C++ requires a bleeding edge compiler. We've been manually triggering these test-deployments in the GitHub UI already. Are you talking about needing to manually trigger when we're ready to release 3.3.4.0rc1? (It would be better to trigger that through a tag/release for the sake of updating to rc2, rc3, etc.)

henryiii: You are stuck with whatever comes on manylinux anyway. This is 2010, so GCC 8 IIRC. The fortran will either already be included (so GCC 8 based) or you will get it from apt, which will then be GCC 4.6 based or something equally as ridiculously old (remember, manylinux2010 is CentOS 6 based).

What error are you looking at?

henryiii: Manually triggering to see if this PR passes. Currently it only builds the tests and not the wheels, so you have to manually trigger GHA to see if the PR can be built into wheels.

henryiii: (which it looks like you are doing, when I look into actions history)

henryiii:
```
checking for CGAL/Exact_predicates_inexact_constructions_kernel.h... no
checking CGAL... no
```
This looks familiar, did we not solve it before?

henryiii: You could try manylinux2014, just to see if CentOS 7 is better, or you could even do manylinux_2_24, which is Debian based but very "new", in linux and pip that are supported, but has an older compiler (GCC 6 IIRC). You'll need the Ubuntu commands for that one.

henryiii:
I've pushed an update using cibuildwheel 2.0.0a4. You can build locally on any computer that has docker and pipx like this:

```bash
CIBW_BUILD='cp39-manylinux_x86_64' pipx run --spec=cibuildwheel==2.0.0a4 cibuildwheel --platform linux
```

I've also added a single wheel to the test matrix. As a test, I've also set manylinux2014, but that fails locally. We could try manylinux_2_24 since we already know the Debian-based procedure to make this work, but that will not be good enough for distribution - not that many Linux machines can run with GLIBC 2.24+. We need to figure out what is missing on CentOS that this needs.

Someone _really_ needs to talk fastjet into cmake. Debugging autotools is ridiculous.

aryan26roy:
Using manylinux_2_24, it was able to build fastjet, but another error pops out. Common advice on the internet is to use "LD_LIBRARY_PATH".

henryiii:
I think you can also pass the path to auditwheel (you can change `[tool.cibuildwheel.linux] repair-command = "..."` if you want). But I would think the paths should have been visible from the build. I'd start by trying it locally and seeing where that .so file is, and look at the module, and see what it links to.

aryan26roy:
I know where that .so file is after the compilation is completed.[https://github.com/scikit-hep/fastjet/blob/9dbbd8b9a3c1df23233b546dff0a7da463bbe2e5/setup.py#L31](url).
This where all the files are supposed to go after fastjet is compiled.

aryan26roy:
https://github.com/scikit-hep/fastjet/blob/9dbbd8b9a3c1df23233b546dff0a7da463bbe2e5/setup.py#L31

aryan26roy:
I know where that .so file is. 
https://github.com/scikit-hep/fastjet/blob/9dbbd8b9a3c1df23233b546dff0a7da463bbe2e5/setup.py#L31
As for what it links to locally :
Screenshot from 2021-06-30 19-22-05

aryan26roy:
@henryiii , it says repair-command not supported in linux section. Am I doing something wrong?

henryiii:
Oops, sorry, it's [`repair-wheel-command`](https://cibuildwheel.readthedocs.io/en/latest/options/#repair-wheel-command)

pyproject.toml
```diff
@@ -6,3 +6,17 @@ requires = [
...
+test-command = "pytest {project}/tests"
+manylinux-x86_64-image = "manylinux_2_24"
+manylinux-i686-image = "manylinux_2_24"
+
+[tool.cibuildwheel.linux]
+before-all = [
+    "apt-get update -y",
+    "apt-get install -y libmpfr-dev libboost-dev",
+]
+repair-wheel-command = "export LD_LIBRARY_PATH=$PWD/_fastjet_core "
```
henryiii, line 22:
```suggestion
repair-wheel-command = [
    "export LD_LIBRARY_PATH=$PWD/_fastjet_core",
    "auditwheel repair -w {dest_dir} {wheel}",
]
```

You need to include the original command, as well - this replaces the command.

henryiii:
```suggestion
repair-wheel-command = "LD_LIBRARY_PATH=$PWD/_fastjet_core auditwheel repair -w {dest_dir} {wheel}"
```

This is probably fine, as well

aryan26roy: Fixed it in the latest commit.

aryan26roy:
@henryiii , even after trying the different paths, it doesn't work. I have assumed that the point of execution of these commands is the same as of setup.py.

henryiii:
I'm not sure $PWD is correct. I think it's going to extract the wheel, find the libraries, then pack it back up. So it would be with relation to where it extracts the libraries - I don't know how to specify this, though.

The correct fix would be to link these with the correct RPath - so the link should be `$RPATH/_fastjet_core/libX.so`, while currently I expect it's actually `$RPATH/libX.so`. Can you do ldd on the python extension module?

aryan26roy:
Here's the ldd on the Python Extension module:
Screenshot from 2021-06-30 21-47-42

henryiii:
For manylinux2014, the boost version is 1.53 (from centOS). CGAL requires 1.66. You could use `boost169-devel` ([from epel](https://centos.pkgs.org/7/epel-x86_64/boost169-1.69.0-2.el7.x86_64.rpm.html)) instead of `boost-devel`. This downloads everything but the kitchen sink (python2, python3, numpy for both, blas, etc), so you probably could eventually just download the parts of this that matter for CGAL. But that still doesn't fix the CGAL discovery:

```console
$ docker run --rm -itv $PWD:/src -w /src quay.io/pypa/manylinux2014_x86_64 bash
# yum update && yum install boost169-devel mpfr-devel
# python3.8 -m pip install -v .
```

(Don't know why I used 3.8 instead of 3.9 there)

henryiii:
`libfastjet.so.0 => not found` is clearly a problem. :)

jpivarski:
Is the Python module supposed to dynamically link to `libfastjet.so` or are the contents of `libfastjet.so` supposed to be compiled into the Python module? I can't remember, but if the design went one way, the `libfastjet.so => not found` should instead be found, and if it went the other way, that line shouldn't be there at all.

aryan26roy:
The path to the lifastjet.so.0 is included in library_dirs in ext_modules in setup.py @jpivarski . I don't know whether that means the files should be found or the line shouldn't be there.

jpivarski:
When you compile your pybind code, does it statically link FastJet into the Python module that creates? If so, then the `_ext` module should not list libfastjet as a dynamic dependency. I _think_ this is what was intended, so many months ago.

If the compilation of your pybind code only dynamically links in FastJet, then it will need to find the libfastjet.so at runtime.

aryan26roy:
Yes, I get that. But locally I run Fastjet just fine with this line being as is.

jpivarski:
Wait—your locally built `_ext` module is saying `libfastjet.so => not found`? Even if it runs, that's a problem. It likely won't run elsewhere with something like this.

aryan26roy:
This is where the compilation happens :
https://github.com/scikit-hep/fastjet/blob/9dbbd8b9a3c1df23233b546dff0a7da463bbe2e5/setup.py#L104

jpivarski:
@henryiii, Aryan reminded me of something that's probably the issue: we dynamically link libfastjet when building the Python module but do not have the path correctly set. As far as the operating system is concerned, it's a broken link. If we were to try to `import fastjet._ext` in Python, it would fail because it wouldn't be able to find libfastjet.so.

However, we do know where the libfastjet.so file is located: it's in the Python package. (Much like Awkward Array, the shared library is copied into the Python package in a customized Install step.) Before executing `import fastjet._ext`, we do

```python
_libfastjet = ctypes.cdll.LoadLibrary(_fastjet_core / "libfastjet.so")
```

where `_fastjet_core` is the known Path. (The use of the "`.so`" extension would be a problem for MacOS, if we were attempting that in this round.) Since this line does a `dlopen` on libfastjet.so, it doesn't attempt to `dlopen` it again to `import fastjet._ext`, and it's able to find all of the symbols that it needs. It works, but it's not a good state of affairs, and surely a wheel audit tool would complain about `libfastjet.so => not found`.

But given that we know exactly where the dynamic library is located, is there a way that we can tell the compiled `_ext` library that information? `LD_LIBRARY_PATH` is not a deployable solution (regardless of whether it works locally), but maybe some RPath thing? (Unfortunately, I don't know enough about library paths to make a good suggestion.)

henryiii:
I think I've finally solved the issue for the _main_ .so, but I think it turns out the fastjet libraries are not built with RPaths, meaning you can't relocate them once they are built. So our library is fine, but the libraries we are bundling fail the audit. :facepalm:

henryiii:
I've committed what I've done as an example, it can be reverted if it's better to do more incrementally; I think we'll have to mess with the generation or patch the .so's after making them from fastjet in the long run. Best solution might be to manually edit the SO's for now.

PR #59, opened 2021-08-04
remove --enable-cgal for rc6, re-enable in rc7

lgray:
https://gitlab.com/fastjet/fastjet/-/blob/fastjet-3.3.4/INSTALL#L55-60

Critically it is necessary to remove this line for builds to work on centos7.
It does not affect building on debian-likes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ class FastJetBuild(setuptools.command.build_ext.build_ext):
             args = [
                 f"--prefix={OUTPUT}",
                 "--enable-allplugins",
+                "--disable-shared",
                 "--enable-cgal-header-only",
                 f"--with-cgaldir={cgal_dir}",
                 "--enable-swig",
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,6 @@ class FastJetBuild(setuptools.command.build_ext.build_ext):
             args = [
                 f"--prefix={OUTPUT}",
                 "--enable-allplugins",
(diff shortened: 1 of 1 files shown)


lgray:
Would recommend that a new rc6 is made!

lgray:
After some discussion with @jpivarski he told me the documentation is wrong and this simply disabled CGAL.

Still, in that case, for keeping things moving along with coffea / coffea-casa, I would prefer a Release Candidate with CGAL simply turned off, and push binary-wheels to rc7.

Coffea casa will be insensitive to this progression of versions.

jpivarski:
We can _temporarily_ accept this, to make an RC release, but earlier experiments showed that removing `--enable-cgal` disables CGAL entirely, even though `--enable-cgal-header-only` is still there. The text of the documentation you highlighted suggests that you only need one option, but we found earlier that you do need both (and cross-checked with Gavin).

We'll eventually need to ship wheels with CGAL (since this is a library and we don't know if it's going to be used in high-multiplicity applications), and will therefore have to turn this option back on. We had some way of verifying that CGAL is, in fact, included, and that was working on GitHub Actions. As long as it does, the binary wheels it builds will be correct. But that's after this next RC release.

@aryan26roy, could you please merge this and make a new RC release? We'll just have to remember to change it back later.

aryan26roy:
@jpivarski , alright I will make a new release with cgal turned off.

PR #65, opened 2021-08-16
Add $ORIGIN/_fastjet_core/lib to RPATH for PyPI wheels.

jpivarski:
Summary of this PR:

   * All ldd "Not Found" errors were resolved by linking with `-rpath=$ORIGIN/_fastjet_core/lib`, with sufficient quoting on the "`$ORIGIN`". Different parts of the configure/make procedure evaluate it at different levels, so just protecting the `$` is not enough: I had to define an environment variable so that `$ORIGIN` would resolve to "`$ORIGIN`".
   * Dropped PxCone because libgfortran is not found. According to @lgray, "PxCone is from 2006 and was most recently cited in a Herwig benchmarking paper for completeness in 2016." It's probably safe to ignore PxCone.
   * Intermediate commits were attempts to get it to work on MacOS. [Last attempt in ci.yml](https://github.com/scikit-hep/fastjet/actions/runs/1137341103), and [last attempt in wheels.yml](https://github.com/scikit-hep/fastjet/actions/runs/1137342116), which are both from bb4afab59cb72f9ea04b520990641402af03a27d.
   * After all that, I tried making [release v3.3.4.0rc7](https://github.com/scikit-hep/fastjet/releases/tag/v3.3.4.0rc7), but this failed because we weren't actually running auditwheel.
   * Restored the cibuildwheel infrastructure to its original state, with `pypa/cibuildwheel@v2.0.1`, following @henryiii's suggestion.
   * FastJet's `./configure`, when it runs inside cibuildwheel, puts the SWIG-generated Python files in a different location. The only way I see to get it is to extract it directly from the `pyinterface/Makefile`. (A banner asking users to set `PYTHONPATH` tells us which variables hold the directory locations.)
   * At the time of writing (just after 1065e43120f5cdf7b3e6e2e292229b355d9fb81a), wheels have been [uploaded to PyPI](https://pypi.org/project/fastjet/3.3.4.0rc8/#files) and successfully tested with `pip install --pre fastjet`!
   * However, the shared libraries have identical files named `libxyz.so`, `libxyz.so.0`, and `libxyz.so.0.0.0`. Presumably, they were originally symbolic links, but now they're duplicates. Removing the two excess copies would reduce the total wheel size by 1/3. (Uncompressed, the extra two copies add up to 52 MB out of 156 MB.) Perhaps the `.0` and `.0.0.0` versions can be dropped in MANIFEST.in?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,9 +31,10 @@ jobs:
       matrix:
         python-version: [3.6, 3.9]
         runs-on: [ubuntu-latest, macos-latest]
-        include:
-        - python-version: pypy-3.6
-          runs-on: ubuntu-latest
+
+        # include:
+        # - python-version: pypy-3.6
+        #   runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,7 @@ jobs:
         CIBW_ARCHS: auto64
         CIBW_BUILD: cp39-*
         CIBW_BUILD_VERBOSITY: 2
+        CIBW_REPAIR_WHEEL_COMMAND: ""
 
     - name: Upload wheels
       uses: actions/upload-artifact@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
       run: python -m pytest -vv -rs -Wd
 
   test_wheels:
-    name: Wheel on linux
+    name: Wheel on Linux
     runs-on: ubuntu-latest
 
     steps:
@@ -68,9 +68,11 @@ jobs:
       with:
         submodules: recursive
 
-    - uses: pypa/cibuildwheel@v2.0.1
+    - uses: pypa/cibuildwheel@v2.1.1
       env:
-        CIBW_BUILD: cp39-manylinux_x86_64
+        CIBW_ARCHS: auto64
+        CIBW_BUILD: cp39-*
+        CIBW_BUILD_VERBOSITY: 2
 
     - name: Upload wheels
       uses: actions/upload-artifact@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.6, 3.9]
-        runs-on: [ubuntu-latest, macos-latest]
+        runs-on: [ubuntu-latest]   # , macos-latest
 
         # include:
         # - python-version: pypy-3.6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,10 +49,6 @@ jobs:
       if: runner.os == 'macOS'
       run: brew install automake swig gmp mpfr boost
 
-    - name: "Try 'xcode-select --install'"
-      if: runner.os == 'macOS'
-      run: xcode-select --install
-
     - name: Install extra deps on Linux
       if: runner.os == 'Linux'
       run: sudo apt-get update && sudo apt-get install -y libboost-dev swig autoconf libtool
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,10 @@ jobs:
       if: runner.os == 'macOS'
       run: brew install automake swig gmp mpfr boost
 
+    - name: "Try 'xcode-select --install'"
+      if: runner.os == 'macOS'
+      run: xcode-select --install
+
     - name: Install extra deps on Linux
       if: runner.os == 'Linux'
       run: sudo apt-get update && sudo apt-get install -y libboost-dev swig autoconf libtool
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,11 +47,11 @@ jobs:
 
     - name: Install compiler tools on macOS
       if: runner.os == 'macOS'
-      run: brew install automake swig gmp mpfr boost gfortran
+      run: brew install automake swig gmp mpfr boost
 
     - name: Install extra deps on Linux
       if: runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get install -y libboost-dev gfortran swig autoconf libtool
+      run: sudo apt-get update && sudo apt-get install -y libboost-dev swig autoconf libtool
 
     #- name: Install dependency
     #  run: python -m pip install awkward
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,6 @@ jobs:
         CIBW_ARCHS: auto64
         CIBW_BUILD: cp39-*
         CIBW_BUILD_VERBOSITY: 2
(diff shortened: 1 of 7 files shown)


.github/workflows/wheels.yml
```diff
@@ -22,27 +22,33 @@ jobs:
...
+        submodules: recursive
+
+    - name: Install compiler tools on macOS
+      if: runner.os == 'macOS'
+      run: brew install automake
 
-    #- uses: pypa/cibuildwheel@v2.0.0a4
-      #env:
-        #CIBW_ARCHS: auto64
+    - uses: pypa/cibuildwheel@v2.0.0a4
```
henryiii, line 43:
```suggestion
    - uses: pypa/cibuildwheel@v2.1.1
```

Is this still manylinux_2_24? Getting it to work on at least 2014 wouldn't likely be very helpful. 2_24 won't run on CentOS 7, for example, too old glibc.

jpivarski: I didn't set that version number.

jpivarski: When 3.3.4.0rc7 is done, I'll commit that suggestion.

jpivarski: It might not be related to this (I might be hijacking this thread), but

```
Binary wheel 'fastjet-3.3.4.0rc7-cp36-cp36m-linux_x86_64.whl' has an unsupported platform tag 'linux_x86_64'.
```

Are we missing a step from some previous dismantling? I'm going to look at the wheels.yml history.

jpivarski:
For this

```
Binary wheel 'fastjet-3.3.4.0rc7-cp36-cp36m-linux_x86_64.whl' has an unsupported platform tag 'linux_x86_64'.
```

upon PyPI upload: it seems that we're missing an `auditwheel repair` step. It looks like wheels.yml never had one?

```bash
auditwheel repair --plat manylinux2010_x86_64 fastjet-3.3.4.0rc7-cp38-cp38-linux_x86_64.whl
```

```
INFO:auditwheel.main_repair:Repairing fastjet-3.3.4.0rc7-cp38-cp38-linux_x86_64.whl
usage: auditwheel [-h] [-V] [-v] command ...
auditwheel: error: cannot repair "fastjet-3.3.4.0rc7-cp38-cp38-linux_x86_64.whl" to "manylinux2010_x86_64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.
```

So the whole build process had to run in the `quay.io/pypa/manylinux2010_x86_64` Docker image, apparently? I thought cibuildwheel would have done this.

jpivarski:
Is it a problem that the cibuildwheel configuration is overriding the default manylinux images?

https://github.com/scikit-hep/fastjet/blob/ea2b05852a57bb15a5f9379376859d08d440dc10/pyproject.toml#L14-L15

I'm going to try removing those lines.

jpivarski:
This is what happened to the repair command: 558bbf2accf5bb2acadbf3c41ae365b4bfe88b26.

.github/workflows/ci.yml
```diff
@@ -66,17 +60,20 @@ jobs:
...
         submodules: recursive
 
-    - uses: pypa/cibuildwheel@v2.0.1
+    - uses: pypa/cibuildwheel@v2.1.1
       env:
-        CIBW_BUILD: cp39-manylinux_x86_64
+        CIBW_ARCHS: auto64
+        CIBW_BUILD: cp39-*
+        CIBW_BUILD_VERBOSITY: 2
+        CIBW_REPAIR_WHEEL_COMMAND: ""
```
henryiii, line 76: Why is this turned off? Wheels should always be "repaired", otherwise they are invalid anywhere other than the machine they were built on.

setup.py
```diff
@@ -89,15 +91,34 @@ def run(self):
...
+            subprocess.check_output(
+                """make -f pyinterface/Makefile --eval='print-pyexecdir:
+\t@echo $(pyexecdir)
+' print-pyexecdir""",
+                shell=True,
+                cwd=FASTJET,
+            )
+            .decode()
+            .strip()
+        )
```
henryiii, line 118:
```suggestion
                cwd=FASTJET,
                universal_newlines=True,
            )
            .strip()
        )
```

You might be able to remove the strip here too, not sure. Python 3.7+ calls this "text=True", but we support 3.6. Also, the recommendation is to use `subprocess.run` for Python 3.5+, but not a big deal here.

(I'd be just fine to merge this in then make improvements like the one above - I'll try to work on the macOS support soon, too)

henryiii (approved):
I'd recommend checking for any duplication and for unhidden symbols  - these wheels are 40MB each, which might be expected, but also might be due to something fixable.

jpivarski:
> I'd recommend checking for any duplication and for unhidden symbols - these wheels are 40MB each, which might be expected, but also might be due to something fixable.

I was thinking about doing something to fix the `libfastjet.so`, `libfastjet.so.0`, `libfastjet.so.0.0.0` duplication (same for all the other libraries), which would reduce 40 MB → 25 MB, before merging.

.github/workflows/ci.yml
jpivarski: It was the last error of the evening: the more extensive wheel-test in `wheels.yml` succeeded, but this single-platform check in `ci.yml` did not. I thought the `wheels.yml` might be running in a different container than the `ci.yml` version. I know from an explicit test that I can't `auditwheel repair` the wheels manually on my system (because of the platform variable).

My first thought was to drop the whole test (the version in `ci.yml`): it's tested in `wheels.yml`. However, `wheels.yml` is opt-in on ordinary builds.

jpivarski:
Formatting (by hand).

.github/workflows/ci.yml
jpivarski: In the message for d4e37649ad4713c15cb5ef5fc0ca9f6c44f0a407, I should have said "reinstating" (by dropping the exclusion).

setup.py
```diff
@@ -89,15 +91,33 @@ def run(self):
...
+        pythondir = pathlib.Path(
+            subprocess.check_output(
+                """make -f pyinterface/Makefile --eval='print-pythondir:
+\t@echo $(pythondir)
+' print-pythondir""",
+                shell=True,
+                cwd=FASTJET,
+            )
+            .decode()
+            .strip()
```
henryiii, line 105:
```suggestion
            subprocess.check_output(
                """make -f pyinterface/Makefile --eval='print-pythondir:
\t@echo $(pythondir)
' print-pythondir""",
                shell=True,
                cwd=FASTJET,
                universal_newlines=True,
            ).strip()
```

.github/workflows/ci.yml
jpivarski: Apparently, it worked. Great!

aryan26roy (approved):

jpivarski:
> > I'd recommend checking for any duplication and for unhidden symbols - these wheels are 40MB each, which might be expected, but also might be due to something fixable.
> 
> I was thinking about doing something to fix the `libfastjet.so`, `libfastjet.so.0`, `libfastjet.so.0.0.0` duplication (same for all the other libraries), which would reduce 40 MB → 25 MB, before merging.

I (or one of us) will remove those duplicates some other time. I'll merge this PR now.

PR #68, opened 2021-08-31
updated fastjet-core

aryan26roy:


chrispap95:
I found what's wrong with FastJet 3.4.0 but I am not 100% sure how to fix it. In `_swig.py` there is this new import:

```python
from _fastjet import FastJetError
```

Changing this to 

```python
if __package__ or "." in __name__:
    from ._fastjet import FastJetError
else:
    from _fastjet import FastJetError
```

fixes it and fastjet is importable again. We could patch this manually although this seems like a very dirty fix. I am not sure how to handle this properly with swig.

chrispap95:
@henryiii this seems to be the culprit https://gitlab.com/fastjet/fastjet/-/blob/master/pyinterface/fastjet.i#L139-142
I can try submitting a PR upstream.

henryiii:
Is there a need for handling the "." in `__name__` case? Looks maybe like holdover from very old Pythons?

Also I think possibly the correct upstream patch is sort of like this:

```
if __name__ == "__main__" and __package__ is None:
    __package__ = "fastjet.wherever.this.is"

from ._fastjet import FastJetError
```

chrispap95:
You are right. This seems unnecessary. I just copied the way swig imports modules and that seems to be kinda old-fashioned.

henryiii:
I would not be surprised if SWIG supports Python 2.5. 🤦

chrispap95:
Nah, the hardcoded requirement is 2.7.0 😂

jpivarski:
Point-zero is oddly specific. (July 4, 2010, apparently.)

jmduarte:
Closed via #122

PR #70, opened 2021-09-03
enabled cgal

aryan26roy:


aryan26roy:
@jpivarski , enabling cgal again breaks the wheel building on linux.

lgray:
So long as fastjet's usage of boost is really header only it seems that: https://github.com/boostorg/boost.git

is plenty fast. But I am guess you tried that and found it lacking?

chrispap95:
ahh actually no! I just used https://www.boost.org/users/download/
I didn't think to look on github! This might be sufficient!

chrispap95:
@lgray did you manage to find any full releases on the boost github page? I only see tags that contain only the main repository (i.e. not recursive). I tried `git clone  --recusive` but the size of the project is huge and it ends up being slower than the official mirrors. I just tried a `curl` from the official mirrors and it took ~4 mins. Not great not super bad. I will make a patch using that for the moment to make sure we have a working version and we can update it if we find a better alternative. 

P.S. Is it legal to host the file ourselves?

lgray:
Hmmm. moment.

lgray:
Ah, I see I should have looked at the repistory structure more closely. That's a bummer.

I think an additional four minutes to build wheels really isn't a big deal in the grand scheme of things, 11m vs 7m doesn't really change the course of one's day. I'd stick with your `curl` patch for now, so go ahead and toss it in.

We can probably host the file ourselves if we want to but that's some rather annoying additional burden to deal with.

lgray:
eeeey it worked!

chrispap95:
Thanks! Looks good now! And the download wasn't bad at all! (I am probably getting a terrible mirror where I am). I guess we are done?!

PR #102, opened 2022-09-23
build wheels for all cpython versions

lgray:
cannot use fastjet with cpython 3.10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
     - uses: pypa/cibuildwheel@v2.1.1
       env:
         CIBW_ARCHS: auto64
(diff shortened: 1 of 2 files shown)


henryiii:
This would slow it down quite a bit by putting everything in one job. This should be added in parallel instead (and we'll need the cibuildwheel update, which we are stuck on).

henryiii:
'3.10' 😳

henryiii:
I've added a "pass" check. We can change the protection settings to look for the pass check, and that way we can change what is required in the future without affecting old/other PRs.

henryiii:
Changed the required check to "pass". Now we can add and remove checks without fiddling with the Actions settings.

PR #105, opened 2022-10-13
Avoid zsh pattern matching in pip command

klieret:


chrispap95:
Hi @klieret, the CI is not working right now. It's a known issue. I have identified the problem and I hope to push a patch asap. Then your merge should be ok.

PR #107, opened 2022-10-14
testing some changes

lgray:


lgray:
going to merge this since residual problems are related to the wheel build as opposed to changes in the vector package.

lgray:
ah - at this point should we just close this?

Issue #98, opened 2022-08-23
Memory corruption for specific chunks


chrispap95:
I am writing this issue to report a problem that I noticed initially happening in my coffea processor but I have now isolated using pure uproot and awkward arrays. The issue is the following: when attempting to retrieve the clustered jets coming from tracks for a specific chunk of a file, I am getting error messages about memory corruption and core dumps. I have prepared the minimum code for reproducing this.

```python
import uproot
import awkward as ak
import numpy as np
import fastjet
import vector

vector.register_awkward()

def main():
        events = uproot.open('qcd.root:Events')

        # Prepare the clean PFCand matched to tracks collection
        Cands = ak.zip(
            {
                "pt": events['PFCands_trkPt'].array(),
                "eta": events['PFCands_trkEta'].array(),
                "phi": events['PFCands_trkPhi'].array(),
                "mass": events['PFCands_mass'].array(),
            },
            with_name="Momentum4D",
        )
        cut = (
            (events['PFCands_fromPV'].array() > 1)
            & (events['PFCands_trkPt'].array() >= 0.7)
            & (abs(events['PFCands_trkEta'].array()) <= 2.5)
            & (abs(events['PFCands_dz'].array()) < 10)
            & (events['PFCands_dzErr'].array() < 0.05)
        )
        Cleaned_cands = Cands[cut]
        Cleaned_cands = ak.packed(Cleaned_cands)

        # The jet clustering part
        jetdef = fastjet.JetDefinition(fastjet.antikt_algorithm, 1.5)
        cluster = fastjet.ClusterSequence(Cleaned_cands, jetdef)

        ak_inclusive_jets = ak.with_name(cluster.inclusive_jets(), "Momentum4D")
        ak_inclusive_cluster = ak.with_name(cluster.constituents(), "Momentum4D")

        return

if __name__ == '__main__':
    main()
```

The file is [here](https://cernbox.cern.ch/index.php/s/yIxqrcqgIMMBlhT). 
My setup is `LCG 102` and fastjet installed on top of that through pip. Specifically: 
```
awkward==1.7.0
fastjet==3.3.4.0rc9
numba==0.51.2
numpy==1.22.3
uproot==4.2.0
vector==0.8.5
```
The error message looks like this:
```
(fastjetEnv) [chpapage@lxplus783 FastjetBug]$ python process.py 
#--------------------------------------------------------------------------
#                         FastJet release 3.3.4
[...]

PR #117, opened 2022-10-23
Fix for invalid memory access issue

chrispap95:
This PR solves #98. My investigation showed that for the last event the line `*ptroff = jets.size()+*(ptroff-1);` is performing an illegal write. 

I am still not sure how `pybind11::buffer_info` allocates memory exactly and why this happens only for some input arrays and not always. However, incrementing the memory allocation by 1 seems to be a quick fix. It is not yet clear to me whether this `+1` needs to be applied to the event contents allocations as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,4 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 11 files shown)


chrispap95:
Actually, this doesn't seem to be the right fix. Will open a new PR.

PR #125, opened 2022-10-26
Introduce clang-format in pre-commit

chrispap95:
`clang-format` in pre-commit for C++ files (only `src/_ext.cpp` for us).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,192 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: None
+AlignConsecutiveMacros: None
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
The [formatted version of _ext.cpp](https://github.com/scikit-hep/fastjet/blob/4ac59d34a0f92bd2e48b5dc70ac22d7a37e6a2e6/src/_ext.cpp) is much easier to read.

That .clang-format file is exactly what I would consider "hard to set up," but having a working model (boost-histogram) helps considerably. I think we'll copy this to format the C++ in Awkward Array. `:)`

henryiii:
> exactly what I would consider "hard to set up,"

It was basically running `-style=llvm`, dumping the values to a file, then modifying them ever so slightly to work nicer on pybind11 and to fix one perceived alignment issue (`int *x` vs `int* x`).

PR #126, opened 2022-10-26
clang-format as pre-commit hook

### LINKED ISSUE #124: Run clang-tidy (or clang-format?) on src/_ext.cpp
There are loops in [src/_ext.cpp](https://github.com/scikit-hep/fastjet/blob/main/src/_ext.cpp) that aren't indented, which would make it a lot easier to follow their logic. If this were Python, we could just run [black](https://black.readthedocs.io/) on all the code (well, if this were Python, the wrong-indentation would be syntactically wrong).

https://github.com/scikit-hep/fastjet/blob/f12e4ea22ac64b2bc9094d1055e8ee8d7da6dd8f/src/_ext.cpp#L134-L146

Is there an equivalent with clang-tidy (or clang-format)?

That is, no meticulous setup, specifying which rules you want to use or not use, just run it and the code is better? Perhaps it would be ideal to have it in CI, but even that is overkill: running it once and getting everything indented once would make a huge impact on readability, even if it isn't enforced for future commits and slowly decays. (Then just run it again!)

jmduarte:
fixes #124 taking config from https://github.com/scikit-hep/boost-histogram (h/t @henryiii)

@jpivarski @chrispap95 does this style look good to you?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,100 @@
+---
+Language:        Cpp
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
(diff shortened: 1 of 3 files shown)


jpivarski:
Note that the branch wasn't deleted (for a closed, unmerged PR).

PR #127, opened 2022-10-31
Fix dangerous realllocs 

chrispap95:
This PR fixes some dangerous `realloc`s introduced with PR #117. The memory is allocated properly using `{len+1}` instead of `{len}`. It is clear that the offsets array needs to be `len+1`. The way this was implemented prior to this PR was by making it `{len}` in the C++ code and then [appending a last element](https://github.com/scikit-hep/fastjet/blob/4c0a3f2c408c3870a47f29da7a54e142fb9ae487/src/fastjet/_multievent.py#L66) in the python code. Instead of that, I suggest that we drop that python line and simply allocate `len+1` in C++. So far the tests pass and everything looks as expected to me. Please let me know if you have any thoughts on this.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,192 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: None
+AlignConsecutiveMacros: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: None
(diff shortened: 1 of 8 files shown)


chrispap95:
Also, I would like to make an `rc2` if this PR is okay since this problem might cause issues to the users.

jpivarski (approved):
This is definitely the right thing to do, including not creating an array of almost the right size and then replacing it with one larger. (It might have been replacing in place, which could account for why it sometimes worked previously.)

jmduarte (approved):
LTGM!

PR #137, opened 2022-11-07
Exclusive jet constituents

rkansal47:
Adding the `exclusive_jets_constituent_index` and `exclusive_jets_constituents` (for a specified number of jets) functions, which retrieve particle constituents of exclusive jets similar to the `constituents` and `constituent_index` functions.

This partially solves #100, but future PRs may want to add the `dcut` arg and implementations for exclusive *subjets* as well.

### DIFF
--- a/src/fastjet/_generalevent.py
+++ b/src/fastjet/_generalevent.py
@@ -630,11 +630,13 @@ class _classgeneralevent:
     def exclusive_jets_constituents(self, njets):
         if njets <= 0:
             raise ValueError("Njets cannot be <= 0")
-        
+
         self._out = []
         self._input_flag = 0
         for i in range(len(self._clusterable_level)):
-            np_results = self._results[i].to_numpy_exclusive_njet_with_constituents(njets)
+            np_results = self._results[i].to_numpy_exclusive_njet_with_constituents(
+                njets
+            )
             off = np.insert(np_results[-1], 0, 0)
             out = ak.Array(
                 ak.layout.ListOffsetArray64(
@@ -678,11 +680,13 @@ class _classgeneralevent:
     def exclusive_jets_constituent_index(self, njets):
         if njets <= 0:
             raise ValueError("Njets cannot be <= 0")
-        
+
         self._out = []
         self._input_flag = 0
         for i in range(len(self._clusterable_level)):
-            np_results = self._results[i].to_numpy_exclusive_njet_with_constituents(njets)
+            np_results = self._results[i].to_numpy_exclusive_njet_with_constituents(
+                njets
+            )
             off = np.insert(np_results[-1], 0, 0)
             out = ak.Array(
                 ak.layout.ListOffsetArray64(
--- a/src/fastjet/_generalevent.py
+++ b/src/fastjet/_generalevent.py
@@ -627,11 +627,62 @@ class _classgeneralevent:
         res = ak.Array(self._replace_multi())
         return res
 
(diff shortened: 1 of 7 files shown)


src/fastjet/_generalevent.py
```diff
@@ -627,6 +627,35 @@ def constituents(self, min_pt):
...
+            )
+            off = np.insert(np_results[-1], 0, 0)
+            out = ak.Array(
+                ak.layout.ListOffsetArray64(
+                    ak.layout.Index64(np_results[0]),
+                    ak.layout.NumpyArray(np_results[1]),
+                ),
+                behavior=self.data.behavior,
+            )
+            outputs_to_inputs = ak.Array(
```
jmduarte, line 648: Up until this point, the basic code of `exclusive_jets_constituents` and `exclusive_jets_constituent_index` is the same. I wonder if `exclusive_jets_constituents` should call `exclusive_jets_constituent_index` or will this cause unnecessary overhead? 

A similar situation applies to `constituent_index` and `constituents`. Clearly, in that case, it was chosen to do it this way. So, this approach is probably the right way to go.

If it turns out it is useful to minimize code duplication, we can deal with it in a follow-up PR.

jmduarte (approved):

jmduarte:
@lgray @chrispap95 thumbs up if this PR looks fine to you. If there are no complaints, I will plan to merge it.

src/fastjet/_generalevent.py
```diff
@@ -627,6 +627,35 @@ def constituents(self, min_pt):
...
+        if njets <= 0:
+            raise ValueError("Njets cannot be <= 0")
+
+        self._out = []
+        self._input_flag = 0
+        for i in range(len(self._clusterable_level)):
+            np_results = self._results[i].to_numpy_exclusive_njet_with_constituents(
+                njets
+            )
+            off = np.insert(np_results[-1], 0, 0)
```
chrispap95, line 640: This numpy operation resembles the code that was causing problems with `inclusive_jets()`. The code was working fine but when tested at scale it was giving seg faults sporadically. In the end, we realized that the offset array needs to be `len+1` in the C++ code and here we just get the array: `off = np_results[-1]`. Now, I realize that I didn't remove all those parts :/ I am fine with keeping this code as is but keep in mind this problem. And we might want to remove all those lines in the future (including the code outside of this PR).

lgray: I think it would be best to sanitize memory access before we merge this. @chrispap95 can you summarize a list of checks / fixes that should be addressed before we merge this?

chrispap95: The fix is to change the allocation size for the offsets array in C++ to `{len+1}` from `{len}` and simplify python code from `off = np.insert(np_results[-1], 0, 0)` to `off = np_results[-1]`. Also, it would be fantastic if this can be done as well for any methods that I forgot to patch. @rkansal47 let me know if you have trouble doing this.

chrispap95:
I think there is a typo `np.results` ~> `np_results`

rkansal47:
Indeed, I was just fixing that :), still some tests failing though.

src/fastjet/_generalevent.py
rkansal47: OK, I believe I've fixed all of them (needed to just insert the 0 in C++ as well).

lgray (approved):

PR #142, opened 2022-11-09
Lund de-clusterings for exclusive jets

rkansal47:
Implementation of the `exclusive_jets_lund_declusterings` method for accessing the Lund plane $\Delta$ and $k_T$ variables per exclusive jet via de-clustering. 

Based on the fastjet-contrib module https://github.com/fdreyer/LundPlane.

Future PRs may want to expand to arbitrary jets and more kinematic variables as needed / requested.


chrispap95 (approved):

lgray:
Should we not also include the calculation of the secondary lund plane if it exists? Or is this a trivial additional operation by the user?

rkansal47:
@lgray I think the PR addresses the main bottleneck from the user side in calculating the Lund plane which is this `while` loop that can’t be parallelised https://github.com/scikit-hep/fastjet/pull/142/files#diff-acd0464ac6e9dddb411598dca843b9c671acb041f047ca8a93f8f12a011baadaR1624-R1631

I think a dedicated function for the secondary plane, or more generally, planes at arbitrary intermediate emissions, would be nice to add in the future, but at least with this they can still be calculated in parallel via awkward array manipulations (e.g. I believe for secondary Lund planes you could run this method on the jets in question with `njets=2` and use the softer subjet’s declusterings).

lgray (approved):

PR #143, opened 2022-11-09
fj-contrib

jmduarte:
- [x] Add `fastjet-contrib` as a submodule
- [x] Compile `fastjet-contrib` as part of `setup.py`as a submodule
- [x] Reuse `LundPlane` code from `fastjet-contrib`
- [x] Check if it runs/passes tests

This approach adds some overhead (compiling all of `fastjet-contrib`), but in principle allows us to add helper functions to use other contributions like SoftDrop, Nsubjettiness, EnergyCorrelationFunctions, etc.

### DIFF
--- a/src/_ext.cpp
+++ b/src/_ext.cpp
@@ -1591,6 +1591,7 @@ PYBIND11_MODULE(_ext, m) {
         }
         jk++;
 
+        auto lund_generator = fastjet::contrib::LundGenerator();
         std::vector<double> Delta_vec;
         std::vector<double> kt_vec;
 
@@ -1618,7 +1619,6 @@ PYBIND11_MODULE(_ext, m) {
           auto prev = ptrjetoffsets[jetidx-1];
 
           for (unsigned int j = 0; j < jets.size(); j++){
-            auto lund_generator = fastjet::contrib::LundGenerator();
             auto lund_result = lund_generator.result(jets[j]);
             auto splittings = lund_result.size();
             for (unsigned int k = 0; k < splittings; k++){
--- a/src/_ext.cpp
+++ b/src/_ext.cpp
@@ -13,6 +13,7 @@
 #include <fastjet/GhostedAreaSpec.hh>
 #include <fastjet/JetDefinition.hh>
 #include <fastjet/PseudoJet.hh>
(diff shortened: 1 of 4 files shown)


src/_ext.cpp
```diff
@@ -1617,17 +1618,12 @@ PYBIND11_MODULE(_ext, m) {
...
-            pair = jets[j];
-            int splittings = 0;
-            while (pair.has_parents(j1, j2)) {
-              if (j1.pt2() < j2.pt2()) std::swap(j1,j2);
-              double Delta = j1.delta_R(j2);
-              Delta_vec.push_back(Delta);
-              kt_vec.push_back(j2.pt() * Delta);
-              pair = j1;
-              splittings++;
+            auto lund_generator = fastjet::contrib::LundGenerator();
```
lgray, line 1621: Would it be possible to instantiate this outside of the double-nested loop?

jmduarte: oops

jmduarte:
```bash
```

jmduarte:
The error occurs when importing `fastjet` so I'm not compiling/linking things appropriately. Maybe @chrispap95 or someone else has a clue on how to fix it?

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/python/cp39-cp39/lib/python3.9/site-packages/fastjet/__init__.py", line 7, in <module>
    import fastjet._ext  # noqa: F401, E402
ImportError: /opt/python/cp39-cp39/lib/python3.9/site-packages/fastjet/_ext.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZTVN7fastjet7contrib13LundGeneratorE
```

chrispap95:
@jmduarte it seems that each contrib is creating a `.a` lib and these libs are not linked against `_ext`. Perhaps, you could add these somehow in`Pybind11Extension(..., library_dirs=[], ... )`. The only problem I see is that these plugins generate static libraries and pybind11 seems to look only for shared libs. @henryiii might know how to solve this.

lgray (approved):

lgray:
We can probably wait to create a new release since the current implementation is correct as far as I can tell. Otherwise this looks fine.

PR #145, opened 2022-11-14
Lund declusterings general event bug fix

rkansal47:


jmduarte (approved):
LGTM! 

Note to self: we should add tests for `_generalevent` methods to prevent this in the future.

PR #156, opened 2022-12-12
fix: update codebase to Awkward 2.0.0.

jpivarski:
I wasn't able to test locally. I _think_ I got all the right changes. Let's see what CI thinks.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    awkward>=1
+    awkward>=2
     numpy>=1.13.3
     vector
 python_requires = >=3.7
--- a/src/fastjet/_generalevent.py
+++ b/src/fastjet/_generalevent.py
@@ -19,7 +19,7 @@ class _classgeneralevent:
         self.multi_layered_listoffset(self.data, ())
         for i in range(len(self._clusterable_level)):
             self._clusterable_level[i] = ak.Array(
-                self._clusterable_level[i].layout.toListOffsetArray64(True)
+                self._clusterable_level[i].layout.to_ListOffsetArray64(True)
             )
             px, py, pz, E, offsets = self.extract_cons(self._clusterable_level[i])
             px = self.correct_byteorder(px)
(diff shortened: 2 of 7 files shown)


jpivarski:
Oh wait—I have to remove debugging code.

setup.cfg
```diff
@@ -31,7 +31,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    awkward>=1
+    awkward>=2
```
jmduarte, line 34: What's the proper way to handle this new version pin?

What I mean is `fastjet<=3.4.0.1` actually only works with `awkward>=1,<2`, but since we already published that, there's no way for pip, etc. to know so effectively all versions of `fastjet<=3.4.0.1` will break if installed with `awkward>=2`.

Or is this not a problem because once we release `fastjet 3.4.0.2`, which will work with `awkward 2.0.0`, everything will be fine?

This is more a meta question not directly related to this PR. But I welcome any guidance.

jpivarski: @henryiii has had a lot to say about this: https://iscinumpy.dev/post/bound-version-constraints/

Applying this case to his arguments, fastjet is a library, to be used as a dependency in someone else's (otherwise version-constrained) application, so it should _not_ be version-capped. Also,

> once we release `fastjet 3.4.0.2`, which will work with `awkward 2.0.0`, everything will be fine

is a good argument. There's a short interval of time in which someone attempting to install will get a combination of packages that don't work together, and that's why I wanted to close the gap as quickly as possible. If someone installs in such a way that they deliberately hold `fastjet` back while allowing `awkward` to update, they'll make problems for themselves. (This is, I think, a summary of the argument against version-capping—adding to it the fact that the "problems they make for themselves" might have been made by one of their dependencies!)

I suppose it's also possible that a dependency management system would upgrade `awkward` without also upgrading `fastjet` without having any positive constraints to do so. I'm quite sure that conda does not work this way, and I'm pretty sure that pip doesn't work this way, or at least not anymore.

jpivarski:
I guess I read the pytest output from Round 1 incorrectly: there was one error causing three test failures. With that error fixed, the other two didn't need to be debugged.

So it's done!

To explain the changes, most of them are just renaming, but in some of them, I replaced a lot of

```python
isinstance(layout, (ak.layouts.ThisArray, ak.layouts.ThatArray))
```

with

```python
layout.is_this or layout.is_that
```

These "`is_*`" booleans are categories of similar array types. All of the integer-sized type variations are gone (e.g. `ListArray64` versus `ListArray32`), which would have simplified the `isinstance` calls on its own, but these "`is_*`" also identify commonly selected groups (e.g. `is_list` is True for `ListArray`, `ListOffsetArray`, and `RegularArray`).

I removed code that depends on now-nonexistent types, `PartitionedArray` and `VirtualArray`. For the tests, however, I just changed the `PartitionedArray` test to work on a concatenated array and the `VirtualArray` test to work on eager data.

I also introduced use of the `copy` method to change one field of a layout node, rather than rebuilding the whole layout node. (If I hadn't done that, I would also have needed to remove references to the `identities`, which have been removed.)

I scanned over the changes once again, and yeah: it was mostly renaming.

setup.cfg
jmduarte: ok great, i was vaguely remembering this post and I'm glad to have been pointed to it again!

jmduarte (approved):
LGTM!

PR #158, opened 2022-12-15
update to v3.4.0.2

jmduarte:
- update to v3.4.0.2; the major change is compatibility with awkward 2.0.0 (the previous version is incompatible and will break)

### DIFF
--- a/src/fastjet/version.py
+++ b/src/fastjet/version.py
@@ -2,7 +2,7 @@
 
 import re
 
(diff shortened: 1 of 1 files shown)


chrispap95 (approved):
Seems fine to me!

lgray:
OK for me - I just need to make sure to not change the pin for the coffea images since v0.7 is incompatible. We'll have awkward2 in v0.8.

PR #159, opened 2022-12-16
Fix for swig 4.1.X

chrispap95:
Swig 4.1.X changed the way it handles functions with `@staticmethod`. Up to version 4.0.2 it used to produce an additional flat-named function for each static method. For example, the function `ClusterSequence.fastjet_banner_stream()` used to also have a flat-name counterpart: `ClusterSequence_fastjet_banner_stream()`. Now, this is gone. 

More info here https://github.com/swig/swig/pull/2137

I just removed those imports. Tests seem to be fine and I couldn't find any uses for them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,9 @@ on:
   pull_request:
   push:
     branches:
-    - master
-    - main
-    - develop
+      - master
+      - main
+      - develop
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


jmduarte (approved):
Looks good to me!

lgray:
Given one of these is some terminal banner for fastjet, check with the authors they don't want that appearing when you load the module or something. Even if it's not in the license it may be a good idea to check? Otherwise these look pretty benign, indeed.

chrispap95:
@lgray I admit I didn't check whether the banner still shows up or not. However, my guess is that it is unrelated to these static methods. I will check and I will reply back here.

chrispap95:
Just verified. Banner shows up as expected.

PR #161, opened 2022-12-16
update to v3.4.0.3

jmduarte:
Anything else we want to include before publishing v3.4.0.3?

It'd be nice to get this out so we can `pip install fastjet` on macOS! 😃


chrispap95:
I would say, let's get those macOS wheels! We can always publish another patch release if needed (not aware of any open issues though).

Issue #165, opened 2022-12-28
Possible to support aarch64 wheels?


matthewfeickert:
:wave: Hi. Would it be possible to support `manylinux2014_aarch64` wheels on PyPI, like [`awkward-cpp` does](https://pypi.org/project/awkward-cpp/5/#files)? As there aren't any available at the moment this means that you need to install from sdist when doing a multi-platform `docker buildx` build (`--platform linux/amd64,linux/arm64`) to provide Docker images that can run without heavy emulation on M1/M2 Macs. I'm currently in the process of trying to do this, and it is a bit trickier than just building for Ubuntu.

lgray:
We'd need a hosted aarch64 runner for that, qemu would take *forever* (x86 builds are 34 minutes!). I am happy to setup a self-hosted aarch64 build machine, I've got some spare cores.

Just to stick a pin in it we should also think about dask_awkward integration.

lgray:
Or we could only spin out wheels for x86/arm64, testing just on arm64 and assume some risk.

lgray:
At the very least I'll give a go to building it on the arm64 machine(s) I've got to see if it completes.

chrispap95:
> Just to stick a pin in it we should also think about dask_awkward integration.

Hi @lgray, how do you imagine dask_awkward integration happening? Do you have any examples from other projects that have already done it?

henryiii:
I believe Cirrus CI has aarch machines.

lgray:
@chrispap95 I think it's not too bad in this case, we could probably even make a decorator to behave differently between `ak` and `dak`. Mostly we need to wrap the various fastjet calls in `dask_awkward.map_partitions` and make sure the predictions of the output (awkward forms) are accurate. 

See examples for regular awkward functionality:
https://github.com/dask-contrib/dask-awkward/pull/133

jmduarte:
Who should we ask about adding Cirrus CI to scikit-hep? 

https://github.com/marketplace/cirrus-ci/order/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW45OTA=?account=scikit-hep

It's free for public repos. So this would be a good way to speed things up for `aarch64` tests/builds.

matthewfeickert:
> Who should we ask about adding Cirrus CI to scikit-hep?


[...]

PR #166, opened 2022-12-30
Test `aarch64` wheels

jmduarte:
Addresses #165.

Building `aarch64` wheels with QEMU does indeed take a while (~1 hour 30 min) as noted by @lgray, so a self-hosted runner or separate CI would be better long term.

<s>Also @rkansal47, there is a minor numerical difference for the Lund clustering that required changing the test from checking for equality to closeness. Interestingly, this is the only test that requires this. You can see the numerical difference in the first failing CI check. I presume this is ok, but do you have any thoughts about why it happens here?</s> I agree with @henryiii that we should probably be using `ak.isclose()` with an appropriate tolerance for all tests anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,16 +78,14 @@ jobs:
           submodules: recursive
 
       - uses: uraimo/run-on-arch-action@v2.5.0
-        name: Build artifact
-        id: build
+        name: Run test in image
         with:
           arch: ${{ matrix.arch }}
           distro: ubuntu_latest
           # Not required, but speeds up builds
           githubToken: ${{ github.token }}
           dockerRunArgs: |
-            --volume "${PWD}:/fastjet"
-          shell: /bin/sh
+            --volume "${PWD}:/fastjet"  --platform linux/arm64/v8
           install: |
             uname -m
             add-apt-repository ppa:deadsnakes/ppa
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,12 +80,15 @@ jobs:
       - uses: uraimo/run-on-arch-action@v2.5.0
         name: Install and test in image
         with:
+          # workaround for https://github.com/uraimo/run-on-arch-action/issues/55
+          env: |
+            GITHUB_WORKFLOW: ${{ github.workflow }}-python${{ matrix.python-version }}
           arch: ${{ matrix.arch }}
           distro: ubuntu_latest
           # Not required, but speeds up builds
           githubToken: ${{ github.token }}
           dockerRunArgs: |
-            --volume "${PWD}:/fastjet" --platform linux/arm64
+            --volume "${PWD}:/fastjet" --platform linux/arm64 --workdir /fastjet
           install: |
             uname -m
             apt-get update
@@ -96,11 +99,11 @@ jobs:
             apt-get update
             apt-get install -y libboost-dev libmpfr-dev swig autoconf libtool python${{ matrix.python-version }} python${{ matrix.python-version }}-dev python${{ matrix.python-version }}-distutils python${{ matrix.python-version }}-venv
             python${{ matrix.python-version }} -m ensurepip
+            echo $PATH
+            echo $PWD
           run: |
             uname -m
             echo $PATH
-            echo $PWD
-            cd /fastjet
             echo $PWD
             python${{ matrix.python-version }} -m pip install '.[test]' -v
             python${{ matrix.python-version }} -m pytest -vv -rs -Wd
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,11 +35,14 @@ jobs:
       matrix:
         python-version: ["3.7", "3.11"]
         runs-on: [ubuntu-latest, macos-latest]
-
-        # include:
-        # - python-version: pypy-3.8
-        #   runs-on: ubuntu-latest
-
+        arch: [x86_64]
+        include:
+          - python-version: "3.7"
+            runs-on: ubuntu-latest
+            arch: arm64
+          - python-version: "3.11"
+            runs-on: ubuntu-latest
+            arch: arm64
     steps:
       - uses: actions/checkout@v3
         with:
@@ -48,6 +51,11 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+
+      - uses: docker/setup-qemu-action@v2.1.0
+        if: matrix.arch != 'x86_64'
+        with:
+         platforms: ${{ matrix.arch }}
 
       - name: Install compiler tools on macOS
         if: runner.os == 'macOS'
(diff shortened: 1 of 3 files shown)


chrispap95 (approved):
I think this is fine as long as it is only a temporary solution!

henryiii:
(All) floats should only be checked for closeness, not equality. Small changes in mathematical operations (such as different hardware choosing floating point operations differently) is fine.

jmduarte:
@matthewfeickert is it easy for you to download the `aarch64` wheel artifact to check it works for your use case?

matthewfeickert:
@jmduarte Yes, I can do that later today US time. I'm currently driving to Texas, so I'm a bit delayed. :)

lgray:
My two cents:
- We care about physics correctness more than we care about compilation succeeding (debugging wise)
- Our problems with fastjet wheels are largely related to the C++ build system for fastjet, rather than the code itself

To me, this means that we should test on arm64 and aarch64, test buildwheel for amd64 only, and build wheel releases for all aarchs. This should keep PR check time down to something semi-reasonable for a while.

matthewfeickert:
> is it easy for you to download the `aarch64` wheel artifact to check it works for your use case?

@jmduarte looks good at first glance! (Downloading and unpacking the artifact under my `/tmp/`)

```console
$ docker run --rm -ti --platform linux/arm64 -v /tmp:/tmp python:3.11 /bin/bash
root@23bbc1041687:/# uname --machine
aarch64
root@23bbc1041687:/# python -m venv venv && . venv/bin/activate
(venv) root@23bbc1041687:/# cp /tmp/fastjet-3.4.0.3-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl .
(venv) root@23bbc1041687:/# python -m pip install fastjet-3.4.0.3-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl 
Processing /fastjet-3.4.0.3-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
Collecting awkward>=2
  Downloading awkward-2.0.4-py3-none-any.whl (541 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 541.3/541.3 kB 7.8 MB/s eta 0:00:00
Collecting numpy>=1.13.3
  Downloading numpy-1.24.1-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl (14.0 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 14.0/14.0 MB 11.1 MB/s eta 0:00:00
Collecting vector
  Downloading vector-0.11.0-py3-none-any.whl (168 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 168.1/168.1 kB 12.0 MB/s eta 0:00:00
Collecting awkward-cpp==5
  Downloading awkward_cpp-5-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl (1.0 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.0/1.0 MB 10.4 MB/s eta 0:00:00
Collecting packaging
  Downloading packaging-22.0-py3-none-any.whl (42 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 42.6/42.6 kB 5.9 MB/s eta 0:00:00
Installing collected packages: packaging, numpy, vector, awkward-cpp, awkward, fastjet
Successfully installed awkward-2.0.4 awkward-cpp-5 fastjet-3.4.0.3 numpy-1.24.1 packaging-22.0 vector-0.11.0
(venv) root@23bbc1041687:/# python
Python 3.11.1 (main, Dec 21 2022, 09:36:17) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import awkward
>>> import fastjet
>>> jetdef = fastjet.JetDefinition(fastjet.antikt_algorithm, 0.6)
>>> 
KeyboardInterrupt
>>> 
(venv) root@23bbc1041687:/#
```

lgray:
throw some random numbers and cluster it!

matthewfeickert:
> throw some random numbers and cluster it!

Updated my previous comment with a replication of the example in the docs.

lgray:
eeww that add-apt-repository line installs the entire universe... let me see if we can do better.

lgray:
what the crap no ensurepip....

lgray:
really into whackamole territory here :-)

lgray:
hooray! that seems to be it!

lgray:
Noooooo, also why is it taking a freaking hour.

jpivarski:
If this is anything like Awkward's ARM build, it takes an hour because the ARM compiler is in a QEMU virtual machine. There was some technical reason for that (@henryiii told me) that the authors hope to get fixed in the future.

.github/workflows/ci.yml
```diff
@@ -67,19 +63,68 @@ jobs:
...
+          arch: ${{ matrix.arch }}
+          distro: ubuntu_latest
+          # Not required, but speeds up builds
+          githubToken: ${{ github.token }}
+          dockerRunArgs: |
+            --volume "${PWD}:/fastjet" --platform linux/arm64
+          install: |
+            uname -m
+            apt-get update
+            apt-get install -y ca-certificates gnupg2 build-essential patch
```
matthewfeickert, line 92:
```suggestion
            apt-get install -y ca-certificates gnupg2 python3-dev build-essential patch
```

> Noooooo

As you're hitting

```pytb
...
    swig_wrap.cpp:178:11: fatal error: Python.h: No such file or directory
      178 | # include <Python.h>
...
```

I assume this means that you need to add `python3-dev` so that the necessary header files are available.

When I successfully built this across both `amd64` and `arm64` earlier in addition to the base environment I had for other software I was building inside of `python:3.9-slim-bullseye`

```dockerfile
RUN apt-get -qq update && \
    apt-get -qq -y install --no-install-recommends \
      gcc \
      g++ \
      gfortran \
      make \
      cmake \
      vim \
      zlib1g \
      zlib1g-dev \
      libbz2-dev \
      rsync \
      bash-completion \
      python3-dev \
      wget \
      ghostscript \
      bc \
      git && \
    apt-get -y clean && \
    apt-get -y autoremove && \
    rm -rf /var/lib/apt/lists/*
```

I also had to add the following

```dockerfile
RUN apt-get -qq update && \
    apt-get -qq -y install --no-install-recommends \
      libboost-dev \
      libmpfr-dev \
      libgmp-dev \
      swig \
      autoconf \
      patch \
      automake \
      libtool && \
    apt-get -y clean && \
    apt-get -y autoremove && \
    rm -rf /var/lib/apt/lists/*
```

so you might need a subset of that. :/

henryiii:
> was some technical reason for that

Cross-compiling is not supported at the moment. Emulation is 5-10x slower than native.

If CC is supported, then it should basically be full speed compile then 5-10x slower test run (still needs to be emulated).

.github/workflows/ci.yml
```diff
@@ -67,19 +63,68 @@ jobs:
...
+            echo 'deb https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu jammy main' >> /etc/apt/sources.list
+            echo 'deb-src https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu jammy main' >> /etc/apt/sources.list
+            apt-key adv --keyserver keyserver.ubuntu.com --recv-keys BA6932366A755776
+            apt-get update
+            apt-get install -y libboost-dev libmpfr-dev swig autoconf libtool python${{ matrix.python-version }} python${{ matrix.python-version }}-dev python${{ matrix.python-version }}-distutils python${{ matrix.python-version }}-venv
```
matthewfeickert, line 97:
```suggestion
            apt-get install -y libboost-dev libmpfr-dev swig autoconf libtool python${{ matrix.python-version }} python3-dev python${{ matrix.python-version }}-distutils python${{ matrix.python-version }}-venv
```

Do you actually need `python3.7-dev`/`python3.11-dev`? Or can you just use `python3-dev` here?

lgray: That will install all of python3.10 as well because of deps. I don't want to get into that kind of dependency hell.

matthewfeickert:
I took the liberty of pushing what should get things passing (feel free to revert and revise as desired).

PR #171, opened 2023-01-20
fix: JetDefinition should not fail on integer argument

### LINKED ISSUE #92: JetDefinition fails with int
If I call `JetDefinition` with an integer, I get the unhelpful error message:
```text
fastjet::Error:  The jet algorithm you requested (2) should be constructed with 1 parameter(s) but was called with 0 parameter(s)
```
Calling with a float causes no such problem. Reproducer below

By the way, this error does not seem to have been properly implemented in python:
```text
---------------------------------------------------------------------------
SystemError                               Traceback (most recent call last)
Input In [7], in <cell line: 1>()
fastjet::Error:  The jet algorithm you requested (2) should be constructed with 1 parameter(s) but was called with 0 parameter(s)

----> 1 jet_def = fastjet.JetDefinition(fastjet.antikt_algorithm, 1)

File ~/miniconda/envs/LHCb-geometry/lib/python3.9/site-packages/fastjet/_swig.py:2793, in JetDefinition.__init__(self, *args)
   2783 def __init__(self, *args):
   2784     r"""
   2785 
   2786     `JetDefinition(JetAlgorithm jet_algorithm_in, double R_in, RecombinationScheme
   (...)
   2791 
   2792     """
-> 2793     _fastjet.JetDefinition_swiginit(self, _fastjet.new_JetDefinition(*args))

SystemError: _PyErr_SetObject: exception <class 'fastjet._swig.Error'> is not a BaseException subclass
```

___
Reproducer:
```python
import fastjet

jet_def = fastjet.JetDefinition(fastjet.antikt_algorithm, 1.)  # no problems
jet_def = fastjet.JetDefinition(fastjet.antikt_algorithm, 1)  # raises error
```

5ean-github:
fixed the error when an int is passed to `JetDefinition()`

Fixes #92


src/fastjet/__init__.py
```diff
@@ -41,7 +41,28 @@
...
+        constructor to fully specify a jet-definition (together with information about
+        how algorithically to run it).
+
+        """
+        if len(args) > 1:
+            args_list = list(args)
+            args_list[1] = float(args[1])
+            args = tuple(args_list)
+
+        _fastjet.JetDefinition_swiginit(self, _fastjet.new_JetDefinition(*args))
```
jmduarte, line 63: Maybe we can try something like the following instead of this line:
```
        super().__init__(self, *args)
```

src/fastjet/__init__.py
```diff
@@ -41,7 +41,28 @@
 from fastjet._swig import GridMedianBackgroundEstimator  # noqa: F401, E402
 from fastjet._swig import IndexedSortHelper  # noqa: F401, E402
 from fastjet._swig import InternalError  # noqa: F401, E402
-from fastjet._swig import JetDefinition  # noqa: F401, E402
+from fastjet._swig import JetDefinition as JetDefinitionNoCast  # noqa: F401, E402
+
+
+class JetDefinition(JetDefinitionNoCast):
```
jmduarte, line 47: Is there any reason to have this up here? or can we move it after all of the imports are done?

jmduarte (changes requested):

PR #189, opened 2023-03-28
add dask_awkward support to fastjet

### LINKED ISSUE #188: dask_awkward support in fastjet
So coffea is getting very close to complete with the awkward2 + dask_awkward upgrade.

It would be great if fastjet would interoperate with dask_awkward easily, so that users don't have to define their own map_partitions calls to access all the functionality of fastjet properly through dask.

I can start a PR with a few examples but I would appreciate some help to make sure we get complete coverage! It shouldn't be that difficult to do.

lgray:
Fixes #188 

Bindings appear to be pretty straightforward at the end of the day.

The trick we use to make delayed operations work is to construct a new ClusterSequence for each concrete array being processed.

Within DaskAwkwardClusterSequence we dispatch the function to be called and the JetDefinition.

There are a large number of functions to convert, but the majority of them seem pretty straightforward.


lgray:
@chrispap95 OK - sorry for the noise here while I beat it into shape. This is all ready for completion now. Let me know if you have any questions! 

There are perhaps a few places where it's not a trivial map_partitions but I haven't thought about it too hard.

lgray:
We probably also need to turn on the thread safety options for fastjet now, since dask's default scheduler is threaded.

lgray:
@chrispap95 ok - finished this up. It would be super useful to try this out with some user analysis and derive some tests before we merge it.

lgray:
@andrzejnovak FYI

chrispap95:
@lgray I can run the SUEP workflow on it. It is using only `inclusive_jets` and `constituents` though. Let me try to set up the environment and I will reply back.

lgray:
@chrispap95 cool - dask-based NanoEvents in the awkward2_dev branch of coffea should work for PFNano so you may just be able to start from that.

lgray:
@jmduarte I know you guys are using some of the lund plane functions and stuff - if you have some time could you give this new software stack a try?

lgray:
There are some bugs in this that need fixing:
- it won't touch data correctly
- it's presently not handling the cases where the function has additional array arguments

lgray:
OK - this is going to need some work now that I've got some tests running on a file.

lgray:
@jpivarski Do you think this is reasonably robust enough? I checked in the code that deeper within all that is ever asked for is E/px/py/pz, but it all looks sorta sketchy IMO.

lgray:
uuughhh

src/fastjet/_pyjet.py
```diff
@@ -228,3 +228,335 @@ def get_child(self, data):
...
+        if ak.backend(array) == "typetracer":
+            array.E.layout._touch_data(recursive=True)
+            array.px.layout._touch_data(recursive=True)
+            array.py.layout._touch_data(recursive=True)
+            array.pz.layout._touch_data(recursive=True)
+            for iarray in arrays:
+                iarray.E.layout._touch_data(recursive=True)
+                iarray.px.layout._touch_data(recursive=True)
+                iarray.py.layout._touch_data(recursive=True)
+                iarray.pz.layout._touch_data(recursive=True)
```
jpivarski, line 249: I was about to say that this isn't sufficient because if the vector is made of a record with fields in a different coordinate system, `iarray.E`, etc. will be newly computed values that do not contain references to the data they were computed from. In that case,

```python
newly_computed_field._touch_data(recursive=True)
```

would do nothing.

_However,_ the act of computation _before_ you get the newly computed field touches the data, so this should be fine. It looks like `tests/test_008-dask.py::test_inclusive_from_file` does that test, and it passed.

So yeah—looks good to me!

jpivarski (approved):

PR #190, opened 2023-03-28
enable thread safety by default in fastjet

lgray:
Turn on thread safety options so users don't run into nasty surprises using dask.


lgray:
This is a side effect of asking for #188

PR #191, opened 2023-03-28
Patch out the override of CXX and CXXFLAGS in fastjet-contrib/Makefile.in

lgray:
This is a blocker in conda.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,5 +4,5 @@ graft fastjet-core
 graft fastjet-contrib
 graft pybind11
 global-exclude .git .gitmodules
-include LICENSE README.md pyproject.toml setup.py setup.cfg patch_fastjet_i.txt patch_clustersequence.txt
+include LICENSE README.md pyproject.toml setup.py setup.cfg patch_fastjet_i.txt patch_clustersequence.txt patch_makefilein.txt
 exclude .cirrus.yml
--- a/patch_makefilein.txt
+++ b/patch_makefilein.txt
@@ -5,7 +5,7 @@ index a4645fd..a7b82e2 100644
 @@ -8,11 +8,6 @@
  SUBDIRS=@CONTRIB_BUILD_LIST@
  SUBDIRS.all=$(SUBDIRS:=.all)
- 
+
 -# these will be overriden if the user specifies CXX or CXXFLAGS with configure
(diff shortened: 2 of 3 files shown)


matthewfeickert:
Related to my confusion in https://github.com/conda-forge/staged-recipes/pull/21052#discussion_r1490312870, it would seem from reading 

```
# these will be overriden if the user specifies CXX or CXXFLAGS with configure
# they will also be overriden by definitions in subsiduary Makefiles
```

(https://github.com/cms-externals/fastjet-contrib/blob/283910e44f2c3c81133fc68c8f4942b9c53da6e3/Makefile.in#L11-L12)

that the [definition of `CXX=g++` in the `Makefile`s](https://github.com/cms-externals/fastjet-contrib/blob/283910e44f2c3c81133fc68c8f4942b9c53da6e3/ScJet/Makefile#L5) of each add-on directory should override this the patching out of `CXX` that this PR adds.

I think this means that I don't understand the build system here though.

lgray:
Oh god it's in each contrib too, good god.

matthewfeickert:
> good god.

Yes, my thoughts exactly. :grimacing:

matthewfeickert:
Can we take advantage of the

```
# global contrib-wide Makefile include may override some of the above
# variables (leading "-" means don't give an error if you can't find
# the file)
-include ../.Makefile.inc
```

(https://github.com/cms-externals/fastjet-contrib/blob/283910e44f2c3c81133fc68c8f4942b9c53da6e3/ScJet/Makefile#L10-L13)

to add in our own `Makefile.inc` that overrides this given that as of 2013 the `Makefile.inc` isn't shipped?

```
2013-02-23  Gavin Salam  <gavin.salam@cern.ch>

	* NEWS:
	* VERSION:
	prepared for 1.001 release

	* contribs.svn:
	added VariableR 1.0.1 and Nsubjettiness 1.0.2  (from Jesse Thaler)

	* configure:
	now outputs version of each contrib (+ other small formatting)

	* Makefile.in:
	distclean now removes Makefile and .Makefile.inc
```

(https://github.com/matthewfeickert/fastjet-contrib/blob/283910e44f2c3c81133fc68c8f4942b9c53da6e3/ChangeLog#L320-L333)

PR #216, opened 2023-06-06
feat: Adding EnergyCorrelator from Contrib

cmoore24-24:
Hello,

I have implemented `exclusive_jets_energy_correlator` to calculate the ECFs and ECFGs of a fatjet, as well as the included combinations (i.e. N2, D2, and so on). 

An important note: I have made the generalized ECF (ECFG) the default case, because the generic ECF does not include baked in normalization, which I suspect most users would expect out of the gate. That said, the combinations will probably be used more often than the the basic ECF, especially NSeries and D2.


chrispap95 (approved):
Looks good to me!

PR #220, opened 2023-06-08
feat: Add exclusive_jets_up_to() interface

### LINKED ISSUE #219: exclusive_jets_up_to()
It would be very useful to expose the functionality of ClusterSequence::exclusive_jets_up_to() in the awkward/multievent interface. I see that exclusive_subjets_up_to is implemented, so I assume it is possible and maybe even easy if that function can be partly copied.

I was prepared to try and implement it myself, but unfortunately I burned all the time I had budgeted for this fighting with the build system. Currently setup.py overwrites CXXFLAGS and LDFLAGS so there is no easy way to pass nonstandard include/lib dirs. That's a pretty easy fix, but the real limitation is fj-contrib's very poor build system, which hardcodes g++ instead of using $(CXX) and doesn't use LDFLAGS at all. At that point I gave up :/

jmduarte:
resolves #219 

@cshimmin

### DIFF
--- a/src/_ext.cpp
+++ b/src/_ext.cpp
@@ -297,11 +297,70 @@ PYBIND11_MODULE(_ext, m) {
             off
           );
       }, "n_jets"_a = 0, R"pbdoc(
-        Retrieves the exclusive jets upto n jets from multievent clustering and converts them to numpy arrays.
+        Retrieves the exclusive n jets from multievent clustering and converts them to numpy arrays.
         Args:
-          min_pt: Minimum jet pt to include. Default: 0.
+          n_jets: Number of exclusive jets. Default: 0.
         Returns:
-          pt, eta, phi, m of inclusive jets.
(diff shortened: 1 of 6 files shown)


jmduarte:
@cshimmin can you try pip installing one of these wheels (non-m1/m2 mac os x or linux): https://github.com/scikit-hep/fastjet/suites/13455142622/artifacts/738101752 to see if you can use `cluster.exclusive_jets_up_to(n_jets)` for your use case?

cshimmin:
Hi Javier,
Thanks, just saw this. Will give it a go tonight and let you know!

Cheers,
-Chase

lgray:
@cshimmin lemme know if you find any problems and I'll approve, codewise it's fine.

lgray:
@cshimmin ping :-)

lgray (approved):

PR #235, opened 2023-07-04
feat: Addition of SoftDrop from the RecursiveTools Contrib

cmoore24-24:
This PR includes the addition of the SoftDrop grooming tool from the RecursiveTools section of fastjet-contrib. The defaults are set to those used in CMS and/or the defaults recommended by the authors of SoftDrop. That said, the options for both the SymmetryCut and RecursionChoice fields have been included. At the moment, the soft drop grooming outputs a record array of the jet constituents, as well as a few values of interest of the jet itself. 

As a small change, EnergyCorrelator has been slightly adjusted, so that the default invocation of the generic ECF class will return the normalized values. The unnormalized values can be returned with a "noramlized=False" argument.


lgray:
@cmoore24-24 can you run pre-commit on this locally and push since somehow your branch protection rules forbid pre-commit.ci from editting this branch?

tests/test_002-exclusive_jets.py
```diff
@@ -211,6 +211,225 @@ def test_exclusive_lund_declustering_multi():
...
+                    atol=0,
+                ),
+            ]
+        )
+    )
+
+    assert ak.all(is_close)
+
+
+def test_exclsuive_jets_softdrop_grooming_multi():
```
lgray, line 319: please add a dask test as well!

cmoore24-24: Can you point me to an example of a dask test I can use as a starting point? I used the other single and multi events to make mine, but I'm not finding other dask tests I can use as a template.

lgray: https://github.com/scikit-hep/fastjet/blob/main/tests/test_008-dask.py#L12

cmoore24-24: I've added a dask test here: https://github.com/cmoore24-24/fastjet/blob/77b5363e526c53d946414f00d7ffcb8fad91f68e/tests/test_002-exclusive_jets.py#L321
Does this look about right?

tests/test_002-exclusive_jets.py
```diff
@@ -4,6 +4,8 @@
 
 import fastjet._pyjet  # noqa: F401
 
+dak = pytest.importorskip("dask_awkward")  # noqa: F401
+distributed = pytest.importorskip("distributed")
```
lgray, line 8: ah! don't do this here - it'll skip the whole test.py file if you are missing dask_awkward / distributed /etc.

Stick this test in the dask tests!

cmoore24-24: Ah sorry about that! I've moved the test to test_008, and reverted test_002

jpivarski:
Just a suggestion: instead of returning a set of filtered jets, you could return an array of booleans. On the Python side, you can use the booleans to slice the original jets and get the filtered jets, but you can also use those booleans for other things. You could, for instance, apply logical operators on groomings with different parameters, to see which jets would be groomed with one set of parameters and not another. That would be harder if you had to identify jets in the output set that are "the same jet."

Also if you do input jets → grooming algorithm → boolean array → slice → filtered jets, the filtered jets share memory with the input jets. It's a view, rather than a copy.

lgray:
This is a very good suggestion - especial for environments like HL-LHC. @cmoore24-24 do you see how to do this?

cmoore24-24:
Is there room for both ideas, potentially? I can definitely see how returning booleans rather than the filtered jets can be helpful when memory is at a premium, but having fastjet do some of the calculation legwork w/r/t the jets is also pretty useful. So maybe keep this implementation as-is, and then down the line have something like `exclusive_jets_softdrop_grooming_slimmed` or some such that returns the booleans. 

As for how to do it, no I'm not really sure off the top of my head. When I have the time to take another look, I can try to figure it out, though.

lgray:
You could have one additional kwarg that's "return_as_mask" or have two functions `exclusive_jets_softdrop_grooming` and `exclusive_jets_softdrop_grooming_mask` depending on what people need.

cmoore24-24:
I've added a way to return a boolean mask that can be applied to a PF candidates array to return filtered jets. As for the implementation; I may be wrong, but I don't believe there's an easy way to make softdrop preserve anything like candidate ID or index (though if there is and I missed it, the feel free to let me know). So the only way I could think to do this was explicit 4 vector matching via nested loops. 
```
for (unsigned int i = 0; i < css.size(); i++){  // iterate through events
          auto jets = css[i]->exclusive_jets(n_jets);
          for (unsigned int j = 0; j < jets.size(); j++){
            auto soft = sd->result(jets[j]);
            nconstituents.push_back(jets[j].constituents().size());
            for (unsigned int k = 0; k < jets[j].constituents().size(); k++){
              for (unsigned int l = 0; l < soft.constituents().size(); l++){
                if (jets[j].constituents()[k].px() == soft.constituents()[l].px() &&
                    jets[j].constituents()[k].py() == soft.constituents()[l].py() &&
                    jets[j].constituents()[k].pz() == soft.constituents()[l].pz() &&
                    jets[j].constituents()[k].E() == soft.constituents()[l].E()){
                  bool_mask.push_back(true);
                  break;
                }
                else if (l == soft.constituents().size()-1) {
                  bool_mask.push_back(false);
                }
              }
            }
          }
        }
```

As a result, it isn't a terribly quick process, but it does work and produce the desired boolean array.

lgray:
We need to do better than quadruple nested loop, this will die in bad ways for HL-LHC analyses. I'll think about a way forward. Would not merge as is.

cmoore24-24:
> We need to do better than quadruple nested loop, this will die in bad ways for HL-LHC analyses. I'll think about a way forward. Would not merge as is.

At this point would it be worthwhile reverting this merge to the pre-mask state and making the addition of the mask a separate PR?

cmoore24-24:
Hello,

I have removed the attempt at adding a mask option for softdrop. That work can serve as a starting point for a separate pull request that adds a mask to softdrop at a later date. The PR is now in its initial state; it returns filtered jet constituents, as well as some attributes of the filtered jets themselves. 

@jmduarte, do you have any comments on the PR as it is now? Can it be merged as is, or are there any remaining changes that need to be made?

Best,
Connor

rkansal47 (approved):
LGTM, I think this is a really useful PR, thanks! I agree the masking can be left for a future PR. I can merge this week unless there are any further comments @jpivarski @jmduarte @lgray.

tests/test_002-exclusive_jets.py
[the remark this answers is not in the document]
cmoore24-24: Looks like I have trouble typing the word "exclusive." Thanks for catching those, fixed.

lgray (approved):

PR #253, opened 2023-09-25
ci: Pin on two digits

rkansal47:
Following Henry's suggestion to reduce updates https://github.com/scikit-hep/fastjet/pull/252#issuecomment-1734198811.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
         with:
           submodules: recursive
 
-      - uses: docker/setup-qemu-action@v3.0
+      - uses: docker/setup-qemu-action@v3
         if: matrix.arch != 'auto64'
         with:
          platforms: all
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
         with:
           submodules: recursive
 
-      - uses: docker/setup-qemu-action@v3.0.0
+      - uses: docker/setup-qemu-action@v3.0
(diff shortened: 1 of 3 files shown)


henryiii:
That’s for cibuildwheel. Some other actions allow 1 digit pinning. It depends on what they publish (usually using the tool from GitHub to make the shortcut).

.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         with:
           submodules: recursive
 
-      - uses: docker/setup-qemu-action@v3.0.0
+      - uses: docker/setup-qemu-action@v3.0
```
henryiii, line 81: Pretty sure this one supports v3

.github/workflows/pr.yml
```diff
@@ -17,6 +17,6 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5.3.0
+      - uses: amannn/action-semantic-pull-request@v5.3
```
henryiii, line 20: Not sure that this one supports.

.github/workflows/wheels.yml
```diff
@@ -99,7 +99,7 @@ jobs:
           name: artifact
           path: dist
 
-      - uses: pypa/gh-action-pypi-publish@v1.8.10
+      - uses: pypa/gh-action-pypi-publish@v1.8
```
henryiii, line 102: This one supports `pypa/gh-action-pypi-publish@release/v1`.

PR #265, opened 2024-01-17
fix: try using flattened array instead of ak.Array(<components>)

lgray:
This appears to be dropping `with_name` information in some cases that causes code using `ak.mask` to fail.

@btovar


lgray:
I don't think this is a bug in fastjet.

lgray:
bug in user code or coffea.

PR #273, opened 2024-02-15
ci: fix failing tests after update to upload-artifact v4

chrispap95:
upload-artifact v4 cannot overwrite artifacts and this causes the tests to fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,8 +98,8 @@ jobs:
       - name: Upload wheels
         uses: actions/upload-artifact@v4
         with:
-          name: wheelhouse-${{ matrix.os }}-${{ matrix.python }}-${{ matrix.arch }}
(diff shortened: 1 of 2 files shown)


chrispap95:
So it fixes the overwrite issue in the CI wheel tests. The real question is if it handles correctly the full wheel building process. Perhaps someone more knowledgeable should take a look at these changes.

lgray:
I personally don't use this action so I'm not sure if this will be correct.

matthewfeickert (commented):
Sorry, I don't have time to give a full useful review right this minute, but I can at least point you to https://github.com/scientific-python/cookie/pull/350 which should show how to simplify things a bit.

Just make sure that both the upload and download are **both** on v4 or you'll have problems.

chrispap95:
I implemented the changes as they are suggested in https://github.com/scientific-python/cookie/pull/350
I quite confident they are okay. I will merge tonight if there no objections.

matthewfeickert (approved):
LGTM. :+1: Things are passing, so I'd say hit it! :rocket:

PR #275, opened 2024-02-17
fix: Add patch to add .Makefile.inc to fastjet-contrib

matthewfeickert:
* As each contribution in `fastjet-contrib` has their own `Makefile` which defines `CXX` and `CXXFLAGS` it is necessary to use a globally used `.Makefile.inc` to change these everywhere. For `fastjet` to be able to control the values from the environment that `fastjet` is being built in, use a `.Makefile.inc` patch to undefine `CXX` and `CXXFLAGS` for `Make`. This allows for the environment varaibles to be picked up.
* Remove `patch_makefilein.txt` as `patch_makefileinc.txt` supreceeds it.

Amends PR https://github.com/scikit-hep/fastjet/pull/191 and is needed for https://github.com/conda-forge/staged-recipes/pull/21052


matthewfeickert:
Thanks for the fast review @lgray!

PR #276, opened 2024-02-17
fix: Correctly use 'undefine' in patch for fastjet-contrib/.Makefile.inc

matthewfeickert:
* The GNU Make undefine directive operates on a variable and not variable assignment.
   - c.f. https://www.gnu.org/software/make/manual/html_node/Undefine-Directive.html

This amends a mistake @matthewfeickert made in PR #275.

Though given https://github.com/conda-forge/staged-recipes/pull/21052#issuecomment-1949796080 I'm not fully clear on if using `undefine` is actually doing what I thought in PR #275.

In https://github.com/cms-externals/fastjet-contrib, all of the contrib directories have `Makefile`s that start the same way:

```
# If you are using this Makefile standalone and fastjet-config is not
# in your path, edit this line to specify the full path
FASTJETCONFIG=fastjet-config
PREFIX=`$(FASTJETCONFIG) --prefix`
CXX=g++
CXXFLAGS= -O3 -Wall -g
install_script = $(SHELL) ../utils/install-sh
check_script = ../utils/check.sh

# global contrib-wide Makefile include may override some of the above
# variables (leading "-" means don't give an error if you can't find
# the file)
-include ../.Makefile.inc
```
(https://github.com/cms-externals/fastjet-contrib/blob/283910e44f2c3c81133fc68c8f4942b9c53da6e3/SoftKiller/Makefile#L1-L13)

So the actions we want to be able to take in the `.Makefile.inc` that we're patching in as `fastjet-contrib/.Makefile.inc` is

1. Unset the values of `CXX` and `CXXFLAGS` from the contrib `Makefile`
2. Set them to be the values that exist in the environment

Given https://github.com/conda-forge/staged-recipes/pull/21052#issuecomment-1949796080 I see that the order here is wrong as the environment variable already exists. So the operations are:

0. The environment `CXX` variable exists.
1. `CXX=g++` in the `Makefile` sets it to `g++`
2. `undefine CXX` fully removes it from the variable scope.

but this is still a problem as we want some value of `CXX` to be defined (the value in the environment).

I don't think what I currently have is the right way to go and I don't think that we can fix this with [conditional assignment either](https://www.gnu.org/software/make/manual/make.html#Conditional-Assignment).

Patching every `Makefile` sounds awful too, so I'd prefer to see all the Makefiles in contrib switch to conditional assingment in the first place.

Example:
```
CXX?=g++
CXXFLAGS?= -O3 -Wall -g
```


matthewfeickert:
Note to future awake self: Use the relevant parts of the patch you wrote in https://github.com/lgray/staged-recipes/pull/2 and then just remove all of

https://github.com/scikit-hep/fastjet/blob/0a70df46a2e8bc7f07b9539decd69f851c820901/setup.py#L104-L109

and [`patch_makefileinc.txt`](https://github.com/scikit-hep/fastjet/blob/0a70df46a2e8bc7f07b9539decd69f851c820901/patch_makefileinc.txt).

matthewfeickert:
@lgray @chrispap95 I found my typo and so this is passing now and ready for review.

chrispap95 (approved):

matthewfeickert:
As this was used in https://github.com/lgray/staged-recipes/pull/2 and approved there I'm going to assume that Lindsey is fine with me removing the patch, as it gets overwritten anyway, and merge this. If not, we can always revert. :+1:

Issue #140, opened 2022-11-07
Re-building from source during development


rkansal47:
If I try to re-build the package from source after making any changes I get the following error: `error: [Errno 17] File exists: 'build/lib.linux-x86_64-cpython-39/fastjet/_fastjet_core'`.

Is there any recommendation for how to re-build after making changes?

And, relatedly, if the changes are only to the Python code or the extension, is there a way to re-build only those components?

chrispap95:
Unfortunately, I don't know a solution to the first question. I usually delete and start from scratch :/ 

For the second part, yes. You can absolutely only recompile `_ext.cpp`. The commands should look like:

```bash
 x86_64-linux-gnu-gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I$(pwd)/src/fastjet/_fastjet_core/include -I$(pwd)/pybind11/include -I$PYTHON_INCLUDE -c src/_ext.cpp -o build/temp.linux-x86_64-3.10/src/_ext.o -fvisibility=hidden -std=c++11
  
  x86_64-linux-gnu-g++ -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-Bsymbolic-functions -g -fwrapv -O2 -Wl,-Bsymbolic-functions -g -fwrapv -O2 -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 build/temp.linux-x86_64-3.10/src/_ext.o -L$(pwd)/src/fastjet/_fastjet_core/lib -Wl,--enable-new-dtags,-R\$ORIGIN/_fastjet_core/lib -lfastjet -lfastjettools -lsiscone -lsiscone_spherical -lfastjetplugins -o build/lib.linux-x86_64-3.10/fastjet/_ext.cpython-310-x86_64-linux-gnu.so
```

jpivarski:
Python builds are not incremental; pip specifically isolates the build directory so that it can't use pre-compiled parts. This error message is saying that's not what's happening here (it's complaining about the `build` directory already existing), so I'm guessing you didn't `pip install .`, but maybe `python setup.py`?

[...]

Issue #279, opened 2024-02-20
Add repository reset tools for developers


matthewfeickert:
After a local build of the the package

```
python -m pip install --upgrade --verbose .
```

the repository is left in a dirty state with additional untracked files in both the repository and the Git submodules. It would be useful to add either a `Makefile` or a `noxfile.py` that impliments `git clean` commands to quickly reset the repository and submodules to a clean state with a single command.

matthewfeickert:
An attempt at this (which hasn't been looked at for a long time, so might need fixing) is the following `build.sh` script

```bash
#!/bin/bash

rm -rf CGAL-5.6*
rm -rf build

cd fastjet-core
git reset --hard fastjet-3.4.2
git clean -f
cd plugins/SISCone/siscone
git clean -f
cd ../../../
cd ..

cd fastjet-contrib
git clean -f
cd ..

python -m pip install --upgrade --verbose .
```

matthewfeickert:
cc @jmduarte and co given Discussion https://github.com/scikit-hep/fastjet/discussions/302

matthewfeickert:
A better script:

```bash
#!/bin/bash

rm -rf CGAL-5.6*

# c.f. https://gist.github.com/nicktoumpelis/11214362
git submodule foreach --recursive git clean -d -f -x
git submodule foreach --recursive git reset --hard

python -m pip install --upgrade --verbose .
```

PR #311, opened 2024-09-11
build: Remove transient directories after install to allow for rebuild

matthewfeickert:
* Addresses part of Issue #279 
* Addresses part of Discussion #302

While doing multiple rebuilds it is advisable to also clean up the state of the Git submodules with `git clean -f`, removing the `OUTPUT` directory at `src/fastjet/_fastjet_core/` and the top level build directory is sufficient to do local development rebuilds &mdash; i.e., the command `python -m pip install --upgrade .` can be run twice in a row without error.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,13 @@ FASTJET_CONTRIB = DIR / "fastjet-contrib"
 PYTHON = DIR / "src/fastjet"
 OUTPUT = PYTHON / "_fastjet_core"
 
+
+# Clean up transient directories to allow for rebuilds during development
+if (DIR / "build").exists():
+    shutil.rmtree(DIR / "build")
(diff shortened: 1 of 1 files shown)


matthewfeickert:
cc @kawaho

matthewfeickert:
@rkansal47 unless you have any thoughts here I'm going to merge this at the end of the day to make @kawaho's development experience easier on PR #312.

rkansal47:
Thanks @matthewfeickert! Does this resolve #140? (Sorry, I'm not able to test the re-build myself currently.)

matthewfeickert:
Ah good catch @rkansal47! I hadn't even noticed that Issue while thinking on Issue #279.

PR #312, opened 2024-09-11
feat: adding groomed jet radius and momentum balance variables for softdrop groomed jet

### LINKED ISSUE #303: Adding groomed jet radius and momentum balance for the soft drop groomed jet
I am proposing to add the groomed jet radius and momentum balance in the return output of the ```exclusive_jets_softdrop_grooming``` function. These two are commonly studied jet substructure variables. The change is simple by basically calling 

```cpp
groomed_jet.structure_of<fastjet::contrib::SoftDrop>().delta_R();
groomed_jet.structure_of<fastjet::contrib::SoftDrop>(). symmetry();
```

and adding the results to the outputs. I am happy to make a pull request if this proposal is accepted.

kawaho:
resolves #303


matthewfeickert:
@kawaho In the future please do not make PRs from your `main` branch and always use a feature branch. Using your `main` branch for PRs makes it impossible to rebase your branch against the upstream without closing the PR.

rkansal47 (changes requested):
Looks pretty good @kawaho!

Unfortunately there is a runtime error with the function now https://github.com/scikit-hep/fastjet/actions/runs/10819914062/job/30018826074?pr=312#step:7:135:

```bash
self = <fastjet._singleevent._classsingleevent object at 0x116f66a90>, njets = 1
beta = 0.0, symmetry_cut = 0.1, symmetry_measure = 'scalar_z', R0 = 0.8
recursion_choice = 'larger_pt', mu_cut = inf

    def exclusive_jets_softdrop_grooming(
        self,
        njets=1,
        beta=0.0,
        symmetry_cut=0.1,
        symmetry_measure="scalar_z",
        R0=0.8,
        recursion_choice="larger_pt",
        # subtractor = 0,
        mu_cut=float("inf"),
    ):
        # if njets <= 0:
        #    raise ValueError("Njets cannot be <= 0")
>       np_results = self._results.to_numpy_softdrop_grooming(
            njets,
            beta,
            symmetry_cut,
            symmetry_measure,
            R0,
            recursion_choice,  # subtractor,
            mu_cut,
        )
E       RuntimeError: std::bad_cast

../../../hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/fastjet/_singleevent.py:215: RuntimeError
```

Perhaps the types for the new variables is incorrect?

Also, can you update the `test_exclusive_jets_softdrop_grooming` function as well please.

kawaho:
@rkansal47 I think the types for the new variables are fine but the error comes from the use of ```dynamic_cast``` in https://github.com/alisw/fastjet/blob/master/fastjet/include/fastjet/PseudoJet.hh#L1126 . It is also strange because the error only happens for macOS, not sure how to resolve it yet....

kawaho:
> @rkansal47 I think the types for the new variables are fine but the error comes from the use of `dynamic_cast` in https://github.com/alisw/fastjet/blob/master/fastjet/include/fastjet/PseudoJet.hh#L1126 

I have done more tests, separating out the ```soft.structure_of<fastjet::contrib::SoftDrop>()``` call,  and can confirm the ```RuntimeError: std::bad_cast``` is from the ```dynamic_cast```. But not sure how to fix yet.

matthewfeickert:
@kawaho I've updated this PR to bring in the state of the upstream `HEAD`, so if you pull your branch will now have the changes from PR #311 in it which should make local development easier. To have a fully clean repository you'll still need to do the `git clean` steps of Issue https://github.com/scikit-hep/fastjet/issues/279, but you should be able to do new builds by just running

```
python -m pip install --upgrade .
```

without _having_ to run the cleaning steps from Issue https://github.com/scikit-hep/fastjet/issues/279. If you have questions just ask here. :+1:

kawaho:
@matthewfeickert Thanks for the PR, it works great.

src/_ext.cpp
```diff
@@ -1703,6 +1705,11 @@ PYBIND11_MODULE(_ext, m) {
             jet_groomed_m.push_back(soft.m());
             jet_groomed_E.push_back(soft.E());
             jet_groomed_pz.push_back(soft.pz());
+
+            auto sd_struct = soft.structure_of<fastjet::contrib::SoftDrop>();
```
kawaho, line 1709: @matthewfeickert and @rkansal47 I would like to hear suggestions on solving this failed test (https://github.com/scikit-hep/fastjet/actions/runs/10819914062/job/30018826074?pr=312#step:7:135). The ```RuntimeError std::bad_cast``` is coming from this specific line calling the dynamic casting in (https://github.com/alisw/fastjet/blob/master/fastjet/include/fastjet/PseudoJet.hh#L1126). The error thrown is happening on MacOS only. I can replicate this error on my own Mac but calling the same function in C++ only script with the same fastjet-contrib version did not give such an error. So I suspect this is related to pybind or how we compile the code. I am unfamiliar with both and would like to hear suggestions for things to try from any experts on this project.

matthewfeickert: Probably want to ask (in addition to @rkansal47) other people that have touched the C++: @jpivarski @jmduarte @chrispap95 @lgray @aryan26roy

lgray:
So - it looks like RTTI in llvm-gcc on macos is more sensitive to consistent const-correctness than pure gcc in linux. From experience with llvm vs. clang this seems to track, but I don't have a definitive answer.

The issue is with FastJet's custom SharedPtr class probably not using `const_cast` in a few places where it should. 
Since we cannot fix that in this PR, I have worked around it using the non-const accessors. I have marked the associated code as dangerous, but we don't do anything exploitative so it should be fine.

lgray:
@kawaho Well, it was eventually figured out. Hopefully this may still be useful.

kawaho:
@lgray thanks! I do not have additional changes. @matthewfeickert can you please review and see if this can be merged.

src/_ext.cpp
```diff
@@ -1690,19 +1694,39 @@ PYBIND11_MODULE(_ext, m) {
...
+              jet_groomed_phi.push_back(soft.phi());
+              jet_groomed_m.push_back(soft.m());
+              jet_groomed_E.push_back(soft.E());
+              jet_groomed_pz.push_back(soft.pz());
+
+              // horrificaly dangerous hack around the fact that
+              // fastjet's custom sharedptr doesn't obey const
+              // correctness and this makes llvm-gcc very sad
+              fastjet::PseudoJetStructureBase* structure_ptr = soft.structure_non_const_ptr();
+              fastjet::contrib::SoftDrop::StructureType* as_sd = (fastjet::contrib::SoftDrop::StructureType*)structure_ptr;
```
rkansal47, line 1715: Claude suggests doing a `dynamic_cast` instead of "C-style" casting. What do you think @lgray?

```
// Safer approach to access SoftDrop structure
fastjet::PseudoJetStructureBase* structure_ptr = soft.structure_non_const_ptr();
if (structure_ptr) {
    auto* as_sd = dynamic_cast<fastjet::contrib::SoftDrop::StructureType*>(structure_ptr);
    if (as_sd) {
        // Successfully cast to SoftDrop structure
        jet_groomed_delta_R.push_back(as_sd->delta_R());
        jet_groomed_symmetry.push_back(as_sd->symmetry());
    } else {
        // Handle case where structure isn't SoftDrop type
        jet_groomed_delta_R.push_back(std::numeric_limits<double>::quiet_NaN());
        jet_groomed_symmetry.push_back(std::numeric_limits<double>::quiet_NaN());
    }
} else {
    // Handle case where structure_ptr is null
    jet_groomed_delta_R.push_back(std::numeric_limits<double>::quiet_NaN());
    jet_groomed_symmetry.push_back(std::numeric_limits<double>::quiet_NaN());
}
```

lgray: Using `dynamic_cast` results in failed tests on macos but not on ubuntu. :-)

As discussed above in the comments, I have to do this specifically because runtime type inference on macos works differently than ubuntu because of const-correctness problems in FastJet itself. All this stems from llvm vs. gcc compiler backends.

So, while the LLM is correct for code that respects standards, we are not dealing with that and have to do this very hacky workaround.

lgray: I'm giving it a quick try on my laptop just in case something was dirty before when I tried it. I would really prefer to use dynamic_cast for the reasons the LLM cited. ;-)

lgray: yeah if I use `dynamic_cast` and the suggested code above I get:
```
Traceback (most recent call last):
  File "/Users/lgray/coffea-dev/fastjet/fail.py", line 137, in <module>
    test_exclusive_jets_softdrop_grooming_multi()
  File "/Users/lgray/coffea-dev/fastjet/fail.py", line 133, in test_exclusive_jets_softdrop_grooming_multi
    assert ak.all(is_close)
           ^^^^^^^^^^^^^^^^
AssertionError
```

From running that test, so it's casting something to a null pointer. :-(
It is, however, clear that the code works and is relatively safe using a more forceful cast. (since the tests here pass on both archs).

Anyway - good try Claude!

PR #313, opened 2024-09-12
build: Use build-system pybind11 over Git submodule

### LINKED ISSUE #309: fastjet is not numpy 2 compatible
All that needs to be done is to move to pybind11 >= 2.12 and to also compile against numpy 2 when building.

matthewfeickert:
Resolves #309 

* Update lower bound on build-system pybind11 to v2.12.0 to add support for compiling for NumPy 2.0.
   - c.f. https://github.com/pybind/pybind11/releases/tag/v2.12.0
* Remove pybind11 submodule to use build-system instead.
* Add an easy to parse log of the environment used for each test, which is useful for debugging differences between CI jobs seperated in time.

For reference, the `pybind11` Git submodule was added in https://github.com/scikit-hep/fastjet/commit/0f9f8fff63761aa63a50471f22880b579b1e6ff2 by @jpivarski, though as this wasn't part of a PR I'm not sure why the Git submodule approach was used over [the other `pybind11` install methods](https://pybind11.readthedocs.io/en/stable/installing.html).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
     "setuptools>=42",
     "setuptools_scm[toml]>=3.4",
-    "pybind11>=2.6.1",
+    "pybind11>=2.12.0",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,9 @@ jobs:
           echo $PATH
           python -m pip install '.[test]' -v
 
+      - name: List installed Python packages
+        run: python -m pip list
+
       - name: Test package
         run: python -m pytest -vv -rs -Wd
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "fastjet"]
 	path = extern/fastjet-core
 	url = https://gitlab.com/fastjet/fastjet.git
-[submodule "pybind11"]
-	path = pybind11
-	url = https://github.com/pybind/pybind11.git
 [submodule "fastjet-contrib"]
 	path = extern/fastjet-contrib
 	url = https://github.com/cms-externals/fastjet-contrib.git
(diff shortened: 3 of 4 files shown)


henryiii:
Why do you have pybind11 and numpy as build requirements? Pybind11 doesn't require numpy at build time.

henryiii:
I don't see any usage of `numpy.h`.

matthewfeickert:
> Why do you have pybind11 and numpy as build requirements?

Error on my part as I didn't read Issue #309 carefully enough and didn't do a `git grep` to check. I'll fix this now. Thanks for pointing this out.

matthewfeickert:
> I don't see any usage of `numpy.h`.
> ...
> Why do you have pybind11 and numpy as build requirements? Pybind11 doesn't require numpy at build time. That's one of it's key selling points, actually.

Ah yes, as you point out

https://github.com/scikit-hep/fastjet/blob/28ec47c78d5042142278cecbc09d1acec0c8a34e/src/_ext.cpp#L20

`pybind11` takes care of that for us.

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
     "setuptools>=42",
     "setuptools_scm[toml]>=3.4",
-    "pybind11>=2.6.1",
+    "pybind11>=2.12.0",
```
matthewfeickert, line 5: I guess this isn't really necessary(?) to bump it given that there's already a lower bound here, but I guess signaling this also doesn't hurt.

matthewfeickert:
@lgray can you comment on the original issues that caused you to open Issue https://github.com/scikit-hep/fastjet/issues/309? I'm not actually clear on what about `fastjet` should be incompatible with NumPy 2.0.

matthewfeickert:
As no one has commented yet, I'm going to merge this myself. As always, PRs approved by a single dev can be reverted as needed by the rest of the dev team.

PR #314, opened 2024-09-13
build: Move Git submodules to extern directory

matthewfeickert:
Move the https://gitlab.com/fastjet/fastjet.git and https://github.com/cms-externals/fastjet-contrib.git Git submodules into an 'extern' directory to make the directory management cleaner.

This implements a comment from @henryiii back in https://github.com/scikit-hep/fastjet/pull/1#issuecomment-820682557.

This does not move the `pybind11` submodule given PR #313 removes it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "fastjet"]
-	path = fastjet-core
+	path = extern/fastjet-core
 	url = https://gitlab.com/fastjet/fastjet.git
 [submodule "pybind11"]
 	path = pybind11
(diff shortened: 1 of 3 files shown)


henryiii:
Personally like calling this directory "extern" (boost-histogram and GooFit use this name)

matthewfeickert:
> Personally like calling this directory "extern" (boost-histogram and GooFit use this name)

@henryiii it is now renamed from "external" to "extern".

matthewfeickert:
Merging as should be uncontroversial, but we can easily revert if needed.

Issue #193, opened 2023-04-12
CGAL explicitly requested and not found or not functional


MycrofD:
Trying to `pip install fastjet` on Mac M2, but getting errors like CGAL not found.

```
      checking CGAL_MAKEFILE... not specified
      checking whether C++ compiler accepts -frounding-math -Werror... no
      checking whether C++ compiler accepts -fp-model strict -Werror... no
      checking CGAL in /private/var/folders/0g/n7p18ft16y3_8bwz00w_1kw81ymyr5/T/pip-install-ksirtcyh/fastjet_17e125686bd34276b7e8111b11946b01/CGAL-5.5.1... lib
      checking for gmp.h... no
      checking for main in -lgmp... no
      configure: error: CGAL explicitly requested and not found or not functional
      This file contains any messages produced by compilers while
      running configure, to aid debugging if configure makes a mistake.

      It was created by FastJet configure 3.4.0, which was
      generated by GNU Autoconf 2.71.  Invocation command line was
```

Steps taken:
1. `brew install cgal`. This installs [CGAL 5.5.2](https://formulae.brew.sh/formula/cgal#default).
2. But [fastjet/setup.py:CGAL_ZIP](https://github.com/scikit-hep/fastjet/blob/39e9ddd689ecbf66af5037192e649bd72dd3712a/setup.py#L25) made me think I might need [CGAL 5.5.1](https://github.com/CGAL/cgal/releases/tag/v5.5.1). So, I tried installing that. 
3. Also tried `pip install --upgrade pip wheel`
4. [Pyjet](https://github.com/scikit-hep/pyjet) seems to install fine without any issue even if it requires `brew install cgal gmp wget`.

But so far nothing helped. Any ideas?

chrispap95:
Unfortunately, we don't have binaries for Apple silicon yet which means that fastjet has to be compiled by pip locally. `pyjet` doesn't compile the C++ FastJet library so it doesn't suffer from these failure cases. I don't have an M1 or M2 machine so I can't offer a lot of help but I believe some of the other devs have machines with Apple silicon so I think it is possible to install fastjet on them. The CGAL that is used here is within the container that pip is creating and, therefore, has nothing to do with what you have in your system. Is this the full error message? In theory, it should have a more detailed printout below. Can you post all of it?

chrispap95:
Did you try after installing the following?
```
brew install make gcc automake swig gmp mpfr boost
```

chrispap95:
[...]

PR #315, opened 2024-09-13
build: Remove manual CGAL install

matthewfeickert:
Addresses (and maybe resolves, will need to check) Issue https://github.com/scikit-hep/fastjet/issues/193.

* build: Remove manual CGAL install
   - Remove the manual download of CGAL from setup.py.
   - Remove `--with-cgaldir` `autogen.sh` flag as CGAL should be installed in a location by the package managers or user that it is findable.

* build: Add manual CGAL install for Linux cibuildwheel 
   - The cibuildwheel Linux container is CentOS 7 based, but CGAL is not packaged for CentOS with `yum` &mdash; only for `apt-get`, Homebrew, and `conda-forge`. For Linux cibuildwheel builds this then requires that the CGAL headers (CGAL is header only) need to be manually downloaded from GitHub and then copied into `/usr/include/`.

* docs: Add `libcgal-dev` install to README 
   - Add `libcgal-dev` to the required packages to install for Linux based development.

* ci: Add cgal install to CI 
   - Use package managers to install `libcgal-dev` (Linux) and `cgal` (macOS) in CI jobs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,13 +15,9 @@ import shutil
 import subprocess
 import sys
 import sysconfig
-import urllib.request
-import zipfile
 
 import setuptools.command.build_ext
 import setuptools.command.install
-
-CGAL_ZIP = "https://github.com/CGAL/cgal/releases/download/v5.6/CGAL-5.6-library.zip"
 
 DIR = pathlib.Path(__file__).parent.resolve()
 FASTJET = DIR / "extern" / "fastjet-core"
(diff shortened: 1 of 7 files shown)


matthewfeickert:
This is ready for review. On my fork I also manually started a wheels build to verify that everything passed across all wheels, and it looks like it will (Linux aarch64 jobs are taking a long time due to emulation)

image

setup.py
```diff
@@ -92,7 +78,6 @@ def build_extensions(self):
                 "--enable-allcxxplugins",
                 "--enable-cgal-header-only",
                 "--enable-cgal",
```
matthewfeickert, line 80:
```suggestion
```

From `./configure --help`:

```
  --enable-cgal           enables link with the CGAL library default=no
  --enable-cgal-header-only   enable build with header-only install of CGAL, e.g. as for CGALv5; in that case do not use --enable-cgal default=no
```

matthewfeickert:
Putting into draft state until PR https://github.com/scikit-hep/fastjet/pull/315 is merged or closed.

lgray:
Closed by #342

PR #326, opened 2024-12-02
Fix typo in definition of constituents (pt instead of p)

chrispap95:
Very trivial change. The input of `constituents()` is supposed to be `min_pt` not `min_p`. When naming the input argument, it causes issues with static code analyzers and type checkers.

### DIFF
--- a/src/fastjet/__init__.py
+++ b/src/fastjet/__init__.py
@@ -347,7 +347,7 @@ class ClusterSequence:  # The super class
 
         raise AssertionError()
 
-    def constituents(self, min_p: float = 0) -> ak.Array:
+    def constituents(self, min_pt: float = 0) -> ak.Array:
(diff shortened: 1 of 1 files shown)


matthewfeickert (approved):

matthewfeickert:
@chrispap95 Thanks for the catch. Assuming this is the only place this occurs, feel free to merge whenever. :+1:

chrispap95:
@matthewfeickert the code runs fine when you call it as: `constituents(min_pt=150)` (this is how we had it in our analysis code). I assume the reason it works is because this function is just a placeholder(?).  I caught it because Pylance was complaining about it. All the other functions are defined with `min_pt` as far as I can tell. Will merge and we can fix any other issues in a separate PR (if found).

PR #332, opened 2025-03-17
build: macos x86_64 wheels using macos-13 runner

lgray:
Let's see if we can get universal2 builds working?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,9 +83,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm] # , macos-latest]
         python: [313]
         arch: [auto64]
+        include:
+          - os: macos-latest
+            python: 313
+            arch: universal2
 
     steps:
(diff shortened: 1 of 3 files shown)


lgray:
nope, even forcing `x86_64` it won't build that arch. :-/

henryiii:
I think it's better to do x86 and arm builds separately, on native hardware, these days. If you want to merge those into a universal2 later, you can.

lgray:
Which runners are x86 these days? macos-13?

henryiii:
Yes, macOS-13 is Intel and macOS-14 is Apple Silicon.

.github/workflows/ci.yml
```diff
@@ -128,14 +124,6 @@ jobs:
...
-      - uses: pypa/cibuildwheel@v2.23
-        if: runner.os == 'macOS'
-        name: "pypa/cibuildwheel@2.23 (macos)"
-        env:
-          CIBW_ARCHS: ${{ matrix.arch }}
-          CIBW_BUILD: cp${{ matrix.python }}-*
-          CIBW_BUILD_VERBOSITY: 2
-
       - uses: pypa/cibuildwheel@v2.23
         if: runner.os != 'macOS'
```
henryiii: ?

Issue #298, opened 2024-06-13
Problems inputting PF candidate array to fastjet clustering


jennetd:
I have a dask_awkward array of particle flow candidates that gives an error when I try to run jet clustering using fastjet. A minimal example is here:
https://github.com/jennetd/hbb-run3/blob/master/data-mc/MinimalExample.ipynb

I am using the newly created image coffeateam/coffea-dask-almalinux8:latest, which includes the following:

```
>>> fastjet.__version__
'3.4.1.3'
>>> awkward.__version__
'2.6.4'
>>> dask_awkward.__version__
'2024.3.0'
```

Please let me know if you have any insights, or if I can provide any more information.

Jennet

jennetd:
I think this error is resulting from the behavior on arrays containing Nones

lgray:
Can you `ak.drop_none` and it suddenly works?

jennetd:
Not exactly. If I flatten the pf candidates it works. I updated the minimal example to show this

PR #333, opened 2025-03-17
fix: propagate metadata (behavior & attrs) in `_generalevent.py`

pfackeldey:
(Hopefully) fixes https://github.com/scikit-hep/fastjet/issues/298

Can you give it a try (@jennetd & @lgray) ?

(Hm... It might be more correct to propagate behavior and attrs in each `self.multi_layered_listoffset` call, but I think it makes no difference here and thus I went with the smaller code change)

### DIFF
--- a/src/fastjet/_generalevent.py
+++ b/src/fastjet/_generalevent.py
@@ -19,7 +19,9 @@ class _classgeneralevent:
         self.multi_layered_listoffset(self.data, ())
         for i in range(len(self._clusterable_level)):
             self._clusterable_level[i] = ak.Array(
-                self._clusterable_level[i].layout.to_ListOffsetArray64(True)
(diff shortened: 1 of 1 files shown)


lgray:
there are a few other places where we should propagate attrs as well!

lgray:
but good catch!

lgray:
Just taking a look at it, `multi_layered_listoffset_input` needs the behaviors and such to be set for its assumptions in calling other functions to be met (if I'm reading this right).

So we should probably propagate there as well.

pfackeldey:
Great, this looks good to me! :+1:

lgray:
The recent changes also fix the problem in the reproducer given.

Will go ahead and merge. I think we're ready for a release after this.

PR #336, opened 2025-03-19
feat: expose contrib::Njettiness

### LINKED ISSUE #286: Add nsubjettiness from fastjet-contrib
Nsubjettiness is one of the most commonly used contrib modules, would be very nice to add an interface to compute them

lgray:
Resolves #286 

There are a few things to clean up in the interface but otherwise does what it says on the label.

default Ns are `[1, 2, 3, 4]`
and the returned list is such that:
```python3
discriminant = returned[..., 1:] / returned[..., :-1]
```


lgray:
@OzAmram @kpedro88 please consider this an assignment for code review.

I would also appreciate a more realistic test with some boosted jets if possible or a simplified prongy jet to check functionality. I could design something but decent simulations of real life are often better.

src/_ext.cpp
```diff
@@ -28,6 +29,58 @@ namespace fj = fastjet;
...
+  OnePass_WTA_KT_Axes,
+  OnePass_WTA_CA_Axes,
+  OnePass_Manual_Axes,
+  MultiPass_Axes,
+  N_AXES_DEFINITIONS
+};
+const std::unordered_map<std::string, MeasureDefinition_t>
+    measure_def_names_to_enum = {
+        {"NormalizedMeasure", NormalizedMeasure},
+        {"UnnormalziedMeasure", UnnormalizedMeasure},
```
kpedro88, line 63: typo Unnormalzied -> Unnormalized

src/fastjet/_generalevent.py
```diff
@@ -708,6 +710,55 @@ def exclusive_jets_softdrop_grooming(
...
+        beta=1.0,
+        R0=0.8,
+        Rcutoff=None,
+        nPass=None,
+        akAxesR0=None,
+    ):
+        if len(njets) == 0:
+            raise ValueError("Must provide at least one njettiness!")
+        if any(njet <= 0 for njet in njets):
+            raise ValueError("Requested njettiness must be > 0!")
```
kpedro88, line 727: s/njettiness/njets/ to be consistent w/ argument naming? (also in the other files)

lgray:
@kpedro88 I was talking to @rkansal47 and it came up that the jet algorithm run ahead of Nsubjettiness is really an open choice. I'm wondering if instead of the current setup I should make Nsubjettiness just take constituents in rather saying that this is Nsubjettiness on inclusive jets for jet definition being used to cluster.

Given how I've named the function, it explain what it does well so I don't think there's presently any lack of clarity, it just implies that if someone wants to run Nsubjettiness on top of an exclusive clustering we'll have to make a PR and add a function.

Thoughts?

kpedro88:
I do think it would be better to be agnostic to the upstream jet clustering choice. It may also be more computationally efficient if the user already has a clustered jet collection and they just want to add Nsubjettiness info.

lgray:
Yeah it's not too heavy a lift either. I'll make that change.

lgray:
OK - with the exception of a realistic test, this is ready to go.

tests/test_002-exclusive_jets.py
```diff
@@ -701,3 +701,49 @@ def test_listoffset_indexed_input():
...
+        ],
+        with_name="Momentum4D",
+    )
+
+    jetdef = fastjet.JetDefinition(fastjet.antikt_algorithm, 0.8)
+    jet_cluster = fastjet._pyjet.AwkwardClusterSequence(array, jetdef)
+
+    constituents = jet_cluster.constituents()
+
+    njettiness_cluster = fastjet._pyjet.AwkwardClusterSequence(constituents, jetdef)
```
kpedro88, line 732: why does `AwkwardClusterSequence()` needs to be called twice?

lgray: The output_wrapper (that contains the input to the jet clustering algorithm in FastJet components) doesn't look like it gets updated between calls, so we're stuck with this funky construction.

i.e. I'm pretty sure I can't do `(jet_cluster.njettiness(jet_cluster.constituents())`

I'll see if I can make it nicer, I agree it's sorta odd. But right now we have some strange internal state to deal with.

I'll poke at it a little to see if something more pleasant can be achieved. OTOH making another ClusterSequence costs nothing.

kpedro88: could the `constituents()` and/or `AwkwardClusterSequence()` calls be made inside `njettiness()`?

lgray: that's how I had it before! the way it's setup now allows us to use an arbitrary jet algorithm before Njettiness.

`.constituents()` implies inclusive, there's a separate constituents for exclusive jets. I will admit it is clunky but it's also outside the scope of this PR to fix it.

I'd vote in favor of flexibility of initial jet clustering algorithm.

kpedro88: > there's a separate constituents for exclusive jets

that was the piece of information I was missing.

lgray: Yeah, indeed, have to make two cluster sequences to set the input arrays to the constituents. It's weird but I think it's an OK price to pay for not having a new function per jet algorithm variant.

OzAmram:
I think a good way to do the test with real jets would be to take one of the new NanoV15 samples ([which now include the PF candidates of the AK8 jets](https://gitlab.cern.ch/cms-nanoAOD/nanoaod-doc/-/wikis/Releases/NanoAODv15)) compute the nsubjettiness values and compare to the centrally computed CMS values 

Unfortunately I don't really have time to set this up at the moment

But there are some TTbar Nano v15 files here if anyone wants to take a stab at it : https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D/TTto4Q_TuneCP5_13p6TeV_powheg-pythia8/RunIII2024Summer24NanoAODv15-150X_mcRun3_2024_realistic_v1-v2/NANOAODSIM

tests/test_002-exclusive_jets.py
kpedro88: What about adding an optional `cluster` argument in `constituents()` that, when true, calls `AwkwardClusterSequence()` internally? (and the same could be done for the exclusive jet version)

lgray: Then we're back to needing a function per jet type, no?
I find that harder to maintain in the long run.

kpedro88: No, I really just mean eliminating one line of code here.

In `_pyjet.py`:
```python
    def constituents(self, min_pt=0, cluster=False):
        result = self._internalrep.constituents(min_pt)
        if cluster:
            return AwkwardClusterSequence(result, self._jetdef)
        else:
            return result
```
then in the test (i.e. in whatever code users will write to use the feature):
```python
constituents_cluster = jet_cluster.constituents(cluster=True)
result = constituents_cluster.njettiness()
```
and a similar pattern can be implemented for the exclusive jet constituents.

lgray: Ah, gotcha, this will require some care with the dask implementation since we have to avoid accidentally sending the client side cluster sequence to a node. It can work.

I would be in favor of this, but it's more substantial than the base scope of this PR. I would take this to an issue and we fix it with a different PR more consistently (this anti-pattern we're unhappy with crops up in other places).

kpedro88: Agreed, opened #337.

lgray:
@OzAmram thanks for the pointers I was able to extract a jet.

lgray:
@rkansal47 When you have time please give this a look over - it's ready to go!

tests/test_002-exclusive_jets.py
rkansal47: Sorry, why doesn't `njettiness` work directly on the first cluster here? i.e. 

```python
jet_cluster = fastjet._pyjet.AwkwardClusterSequence(array, jetdef)
njetti = jet_cluster.njettiness()
```

I'm confused because that seems to work on the "realistic" test below without calling `constituents()` first?

lgray: If we called njettiness on the input data it would do the forced exclusive clustering you do in njettiness to the whole event!

lgray: So you have to cluster jets first then do njettiness on the constituents of the jets l.

rkansal47: Ah so that would give the event `njettiness` rather than per-jet `nsubjettiness`? OK then, LGTM, thanks!

rkansal47 (approved):

PR #340, opened 2025-04-07
build: windows wheels

lgray:
Enable windows builds of fastjet wheels, pointing to in-progress MRs for fastjet and fastjet-contrib.

c.f. for in-progress MRs:
- [x] https://gitlab.com/fastjet/fastjet/-/merge_requests/25
- [x] https://gitlab.com/fastjet/siscone/-/merge_requests/15
- [ ] https://github.com/cms-externals/fastjet-contrib/pull/19

We'll merge this once all that is in releases, except for contrib...

### Below is historical, documenting progress towards #342 
This branch will be where we figure out `pip install .` and `cibuildwheel` for scikit-hep/fastjet using the in-progress cmake build branches for fastjet/siscone/fastjet-contrib.

I need to get the fastjet-contrib edits into a github repository so we can submodule it correctly. 
Right now I have it working locally from the fastjet-contrib svn repository + patches.

At least on my laptop I can get the `scikit-hep/fastjet` extension compiling and linking.
It doesn't yet function at runtime, but it's only a matter of time.

I'll send some further notifications when it's ready for a more serious look.

**One thing we need to solve:**
- [x] MACOSX_DEPLOYMENT_TARGET=11.0 ~(instead of 14.0/13.0 which is what gets us all green right now)~
- [x] Reduce windows wheel build time (presently 5-7m full build time ~40m-1hr installing dependencies~)
- [x] Reduce windows wheel build size (presently ~16MB~ ~4.3MB~ ~2.8MB~ 2.4MB / wheel in windows, ~2.0MB in other OSes)

@matthewfeickert @henryiii


lgray:
OK - the present setup builds on my mac using the following commands:

```
git clone git@github.com:scikit-hep/fastjet.git -b cmake-build-work --recursive
cd fastjet
mkdir build
cmake -S . -B build -DCMAKE_BUILD_TYPE="Release" -DFASTJET_ENABLE_PYTHON=ON -DFASTJET_ENABLE_CGAL=ON -DFASTJET_PYTHON_PACKAGE_NAME=fastjet_cxx
cmake --build build --clean-first --parallel
cd build
python -c 'import _ext; print(_ext)'
```

Depending on your circumstances you may need to specify `pybind11_DIR`.

I'm gonna leave it at that for now y'all can feel free to poke around on the PR with additions, etc.

lgray:
OK - cool so now it's at least compiling (but not finishing the install!) with scikit-build-core. CGAL is on in all the cases, installed via system libs.

The errors I'm seeing don't make a lot of sense, so would now definitely appreciate help in getting the CI green again :D

lgray:
hmmm it would be nice to be able to browse the packages available in manylinux online.

anyway leaving it here for now. more than decent progress.

lgray:
@henryiii 

```
git clone git@github.com:scikit-hep/fastjet.git -b cmake-build-work --recursive
cd fastjet
pip install . -vv
python -c 'import fastjet; print(fastjet)'
```

lgray:
I'm also not sure what's up with this metadata error in the CI install. This doesn't happen locally.

pyproject.toml
```diff
@@ -23,3 +34,20 @@ before-all = [
 ]
 # Skip musllinux builds for the moment
 skip = "*-musllinux_*"
+
+[tool.scikit-build]
+metadata.version.provider = "scikit_build_core.metadata.setuptools_scm"
+cmake.version = ">=3.11.0"
```
henryiii, line 40: 3.15 is the first version scikit-build-core supports.

pyproject.toml
```diff
@@ -23,3 +34,20 @@ before-all = [
...
+sdist.include = ["src/fastjet/version.py"]
+wheel.install-dir = "fastjet"
+
+[tool.scikit-build.cmake.define]
+FASTJET_ENABLE_PYTHON = "ON"
+FASTJET_ENABLE_CGAL = "ON"
+FASTJET_PYTHON_PACKAGE_NAME = "_swig"
+FASTJET_CUSTOM_PYTHON_INSTALL = "./"
+CMAKE_BUILD_TYPE = "Release"
```
henryiii, line 49: You don't need this, it's done by scikit-build-core, unless you are building with a debug Python. And it's got it's own scikit-build-core setting.

pyproject.toml
```diff
@@ -1,10 +1,21 @@
 [build-system]
 requires = [
-    "setuptools>=42",
-    "setuptools_scm[toml]>=3.4",
-    "pybind11>=2.12.0",
+  "setuptools_scm[toml]>=3.4",
```
henryiii, line 3: Not needed, scikit-build-core will request it via the plugin.

CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+cmake_minimum_required(VERSION 3.11)
```
henryiii, line 1:
```suggestion
cmake_minimum_required(VERSION 3.15...3.26)
```

(or 4.0 if you set `PYBIND11_FINDPYTHON ON`)

pyproject.toml
```diff
@@ -23,3 +34,20 @@ before-all = [
 ]
 # Skip musllinux builds for the moment
 skip = "*-musllinux_*"
+
+[tool.scikit-build]
```
henryiii, line 38:
```suggestion
[tool.scikit-build]
minimum-version = "build-system.requires"
```

And I'd probably make that 0.10 or 0.11.

pyproject.toml
henryiii:
```suggestion
cmake.version = "CMakeLists.txt"
```

If you set the min correctly in CMakeLists.txt.

lgray:
Nice, only the macos wheels are failing due to what looks like overlinking?
```
delocate.libsana.DelocationError: Already planning to copy library with same basename as: libgmp.10.dylib
```

Still, extremely positive and especially given how much this has simplified the installing and wheel building process.

lgray:
@gavinsalam FYI this is what is driving changes in the draft PRs over in fastjet/siscone. Once we get this building wheels in all cases then requirements are satisfied on my side. As you can see it's pretty close!

So long as we're careful to reconcile your requirements with the requirements here we shouldn't run into too much trouble. This will also have a side effect of making the new fastjet/siscone/contrib cmake setup fairly robust; cibuildwheel is picky.

lgray:
Now that that particularly angry yak is shaved. Let's clean this up and ship it. :-)

lgray:
@gavinsalam Thanks for the info! It's not CGAL that's causing problems here (though it does take a long time to install in windows, I have some ideas to mitigate that), it's weirdly strict standards compliance from MSVC + fastjet-contrib not having been completely vetted.

With a little more fiddling it should be working! Only two compilation errors left that I see. The only unfortunate thing is the hour long build time right now, but since I have it working I don't want to now remove CGAL. 😒

lgray:
Cool - windows tests pass.

Now to shoehorn it into cibuildwheel.

lgray:
@matthewfeickert @henryiii I've got the windows wheel matrix fully up and running. I've added two tasks that I would appreciate help on now that we've got working recipes. Any help you is deeply appreciated in these directions, it'll vastly help with maintainability.

lgray:
A quick check of turning off debug info in the windows wheels brought it down to 4.5MB on my home desktop. That's reasonable already. I'll put it in the full wheel matrix if everything works out in the short-form CI.

lgray:
looks like things are failing due to gitlab getting angry. I'll start 'em again tomorrow.

lgray:
Not too bad, down to 7 minutes for the windows build now.

Will propagate it later to the wheels.

lgray:
Nice! The dll build allows us to get to 2.4MB wheels on windows, which is commensurate with the other OS's wheel sizes. 

I think this is now very well under control.

lgray:
OK - there we go now we ship the complete fastjet-contrib uni-library. All plugins except for pxcone are enabled. It's more or less fully featured and ~future proof.

lgray:
OK - this latest change is using a top-level only CMake setup for fastjet-contrib. Somehow this has broken sdist, @matthewfeickert @henryiii a look would be appreciated.

I should be able to converge on a nice patchset for fastjet-contrib at this point.

The unfortunate thing is getting all the variants working does require code changes in fastjet-contrib, but they're relatively minimal.

lgray:
@gavinsalam You may find the top-level-only cmake implementation for fastjet-contrib here:
- https://github.com/lgray/fastjet-contrib/pull/1

Right now it includes the necessary patches for windows. It's not actually required to apply them for unix.

lgray:
Unsurprising, but there are no numpy windows arm64 wheels yet. Fastjet itself actually finished compiling though, we'll be ready when it's available.

henryiii:
You can add the Rust toolchain (see https://github.com/actions/partner-runner-images/issues/77), and I think it's coming soon anyway. I'd skip older Python's on Win ARM, there really aren't legacy users there anyway.

lgray:
Yikes, now it dies on building pyarrow.

lgray:
What I've decided to do in the mean time is just skip tests on windows-11-arm64. So much of the infrastructure is missing or won't build at the moment, and the library itself builds fine and tests fine in x64 windows and arm64 linux/macos. So we're very likely safe.

lgray:
This is now waiting for the release of fastjet-3.5.1. I'm going to flip it out of draft mode.

lgray:
OK - this is now using fastjet-3.5.1

lgray:
@gavinsalam I shall require your blessing to merge this. :-)

I think the only point of concern is using the cmake patchset atop fastjet-contrib, but it is localized entirely to the build step of this project. It's not presenting it to users for wider consumption, so it's much less of a risk.

matthewfeickert (approved):

PR #345, opened 2025-06-11
ci: update to cibuildwheel 3.0

henryiii:


pyproject.toml
```diff
@@ -90,11 +90,11 @@ strict_equality = true
 ignore_missing_imports = true
 
 [tool.cibuildwheel]
-skip = "pp*"
+# Skip musllinux builds for the moment
+skip = "*-musllinux_*"
 test-extras = "test"
 test-command = "pytest {project}/tests"
 manylinux-x86_64-image = "manylinux2014"
-manylinux-i686-image = "manylinux2014"
```
henryiii, line 97: Do you need 32-bit linux? We've removed it from `"auto"`, so you need to add `"auto32"` for it. I don't think numpy ships 32-bit linux wheels?

lgray: We can also now try building the musl linux variant FWIW

pyproject.toml
```diff
@@ -90,11 +90,11 @@ strict_equality = true
 ignore_missing_imports = true
 
 [tool.cibuildwheel]
-skip = "pp*"
+# Skip musllinux builds for the moment
+skip = "*-musllinux_*"
 test-extras = "test"
 test-command = "pytest {project}/tests"
 manylinux-x86_64-image = "manylinux2014"
```
henryiii, line 97: Why are you pinning an old version here? We've moved to manylinux_2_28, and numpy 2.3 only ships 2_28 wheels. Why no numpy 2.3, by the way?

lgray: There's a bug in awkward array that makes it incompatible with 2.3 for the time being.

henryiii: You can leave 3.9 and 3.10 as manylinux 2014 if you want, that's also an option with overrides.

lgray: I don't think there's a very good reason to.

PR #385, opened 2026-04-16
build: add 3.14t testing and wheels

lgray:
This does not make this package threadsafe, and does not mark the module as such.
This is only to provide wheels.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.14"]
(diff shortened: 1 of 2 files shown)


henryiii:
`PYBIND11_MODULE(_ext, m)` -> `PYBIND11_MODULE(_ext, m, py::mod_gil_not_used())`

lgray:
I thought that was *only* if the module is actually threadsafe? I am not sure if this one is.

henryiii:
That will keep this module from locking the GIL for everyone that loads it.

It'll work fine from one thread. Thread safely issues happen if multiple threads use this, that would be a good thing to check next.

Use

```
Programming Language :: Python :: Free Threading :: 1 - Unstable
```

PR #386, opened 2026-04-16
build: drop python 3.9 from wheels and testing

lgray:


henryiii:
I'd recommend `ruff` -> `ruff-check` in pre-commit, and using `ruff-format` instead of black.

lgray:
Sure we can get that in with this change.

PR #389, opened 2026-05-12
fix: Fixing broken ECF function in _generalevent.py

cmoore24-24:
Hello! 

I was alerted to a few issues with the way the ECF function was constructed in `_generalevent.py`. To be specific:

1. None of the arguments were being passed to C++, so it would return the defaults in `_ext.cpp`
2. Input arguments in the wrong order
3. C++ returns an array (size and data), and `_generalevent.py` previously returned just the first element.

All of these are fixed in this pull request. As a small extra bit, The DaskAwkward cluster sequence previously had `normalized=False` in `_pyjet.py`, while other arguments for normalized were all set to `True`, so I updated that for consistency.

### DIFF
--- a/src/fastjet/_pyjet.py
+++ b/src/fastjet/_pyjet.py
@@ -549,7 +549,7 @@ class DaskAwkwardClusterSequence(ClusterSequence):
         angles=-1,
         alpha=0,
         func="generalized",
-        normalized=False,
+        normalized=True,
     ):
         return _dak_dispatch(
             self,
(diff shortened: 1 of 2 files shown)


lgray:
@cmoore24-24 since we ran into problems with these functions could you please add some tests to make sure things don't break in the future? Thanks!

src/fastjet/_pyjet.py
```diff
@@ -549,7 +549,7 @@ def exclusive_jets_energy_correlator(
         angles=-1,
         alpha=0,
         func="generalized",
-        normalized=False,
+        normalized=True,
```
lgray, line 552: You probably want to make this change over all of the various interface classes. There would be a few places to change this default.

